PR #5, opened 2020-12-19
add gitattributes to reduce composer package size 

Sysix:
adding all files to .gitattributes, which the composer package don't need.

Should always include:
- Pipline/Github Actions
- Tests Files
- Docs
- etc.


damcclean:
Thanks! I always wondered what that file was for, now I know.

Sysix:
you can do much more with it, but for the beginning it is enough :)

Issue #23, opened 2021-04-17
Live Preview support


duncanmcclean:
## What would you like to see added?

Lots of people recently have been asking for Runway to add support for Live Preview.

PR #28, opened 2021-05-01
Front-end routing

### LINKED ISSUE #25: Front-end routing
## What would you like to see added?

I'd like for Runway models to provide their own front-end routes so they act just like collections. 

We may be able to achieve this by adding our own repository to this list... https://github.com/statamic/cms/blob/3fab3f0d3b8600da6e597e6a8bb0d63e70502e21/src/Providers/AppServiceProvider.php#L106

^ that list is used for the `Data` facade which is whats used for content lookups in the `FrontendController`, so in theory adding a custom repo in there and some other bits and bobs should get us pretty close to a prototype. 

If this is something you'd like to see added, please leave a 👍 reaction to this issue.

duncanmcclean:
## Description

This pull request introduces front-end routing functionality as part of Runway. Essentially, it hooks into the `DataRepository`, the same way entries/terms/etc do.

Eventually, we'll use this foundation to implement live previews (#23).

**Example**

Say you have an e-commerce store and you're storing your products in the database. You probably at some point want to display those products on their own pages. Maybe something like `/products/slug-of-product`. 

Basically, this feature will allow you to do that. To use it, there's two things you need to do:

First, add a `route` to your model's configuration:

```php
// config/runway.php

'route' => '/products/{{ slug }}',
```

In addition, you'll need to add the `HasRoutes` trait and the `Responsable` interface to your Eloquent model.

```php
// app/Models/Product.php

use DoubleThreeDigital\Runway\Routing\Traits\RunwayRoutes;
use Illuminate\Contracts\Support\Responsable;

class Product extends Model implements Responsable
{
    use RunwayRoutes;
```

**To Do**

* [X] Don't rebind the whole `DataRepository` ourselves, extend on what's already there.
* [X] Implement a table based system for URIs
* [X] Add command to manually rebuild URI cache
* [ ] Ensure variables are augmented for use in Antlers #32
* [x] Add a 'Visit' button when editing models in the Control Panel
* [x] Do we need the user to add the `Responsable` interface to their models, can we do that in our trait?
* [x] Ensure templates/layouts are configurable
* [x] Write tests
* [x] Write documentation

## Related Issues

Closes #25.


src/Routing/Traits/RunwayRoutes.php
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    public function toResponse($request)
+    {
+        return (new ResourceResponse($this))->toResponse($request);
+    }
+
+    public function template()
+    {
+        return 'default';
```
duncanmcclean, line 40: Need to ensure this is configurable.

src/Routing/Traits/RunwayRoutes.php
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    public function layout()
+    {
+        return 'layout';
```
duncanmcclean, line 45: Need to ensure this is configurable.

src/Routing/Traits/RunwayRoutes.php
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    public function route()
+    {
+        return $this->runwayUri->uri;
+    }
+
+    public function routeData()
+    {
+        return [];
```
duncanmcclean, line 25: This should probably return something more than an empty array. The question is: what should that be?

src/Routing/ResourceResponse.php
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    protected function cascade()
+    {
+        return Cascade::instance()->withContent($this->data)->hydrate();
+    }
+
+    protected function adjustResponseType()
+    {
+        // $contentType = $this->data->get('content_type', 'html');
```
duncanmcclean, line 89: Should we just get rid of this method altogether or make it work if a `content_type` field exists on the model?

src/Routing/ResourceResponse.php
```diff
@@ -0,0 +1,158 @@
...
+        // if ($contentType !== 'html') {
+        //     $this->headers['Content-Type'] = self::contentType($contentType);
+        // }
+
+        return $this;
+    }
+
+    protected function addContentHeaders()
+    {
+        // foreach ($this->data->get('headers', []) as $header => $value) {
```
duncanmcclean, line 100: Same thing here. Configurable or grab from model?

PR #47, opened 2021-06-30
Adds nexttick for dirty state avoidance.

johncarter-:
## Description

Fixes the dirty state warning on create new model.

## Related Issues

https://github.com/doublethreedigital/runway/issues/41

### DIFF
--- a/resources/js/components/Publish/PublishForm.vue
+++ b/resources/js/components/Publish/PublishForm.vue
@@ -119,7 +119,7 @@ export default {
                     this.$refs.container.saved()
 
                     if (this.isCreating) {
-                        window.location.href = response.data.redirect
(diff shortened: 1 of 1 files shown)


resources/js/components/Publish/PublishForm.vue
```diff
@@ -119,7 +119,7 @@ export default {
                     this.$refs.container.saved()
 
                     if (this.isCreating) {
-                        window.location.href = response.data.redirect
+                        this.$nextTick(() => window.location.href = response.data.redirect);
```
duncanmcclean, line 122: Could you do something like this instead? Sorry... 

```js
this.$nextTick(() => {
    window.location.href = response.data.redirect
});
```

duncanmcclean (approved):

PR #50, opened 2021-07-01
Adds support for save key bindings on publish form

johncarter-:
## Description

Adds mod+s, mod+return key bindings for saving models.

## Related Issues

Adds https://github.com/doublethreedigital/runway/issues/42


duncanmcclean:
Thanks!

PR #59, opened 2021-07-28
Fix prop issue in ListingVue

Skullbock:
## Description

Listing config should be an object, since it's used as one.


duncanmcclean (approved):

duncanmcclean:
Thanks!

PR #60, opened 2021-07-28
Use date field's own format to edit the resource

Skullbock:
## Description

When editing a Date field, let's use the user defined format if set, or a sensible default if not

### DIFF
--- a/src/AugmentedRecord.php
+++ b/src/AugmentedRecord.php
@@ -19,10 +19,9 @@ class AugmentedRecord
                 $value = $record->{$key} ?? $value;
 
                 if ($value instanceof CarbonInterface) {
-                    $field = $resource->blueprint()->field($key);
-
                     $format = $defaultFormat = 'Y-m-d H:i';
-                    if ($field) {
+
+                    if ($field = $resource->blueprint()->field($key)) {
                         $format = $field->get('format', $defaultFormat);
                     }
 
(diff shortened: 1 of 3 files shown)


duncanmcclean:
I'm not sure I completely understand what the issue is here - do you have an example?

(thanks for these PRs by the way, I really appreciate it!)

Skullbock:
This fixes 2 issues:
1. The default date format was laking the `:s`. This meant that any datetime stored in mysql would fail because carbon was trying to parse `Y-m-d H:i` and the date is stored as `Y-m-d H:i:s`
2. While i was at it, since each date field in statamic allows for a particular date format to be used, i also thought to "pick that up" from the config and use it. A use case could be a date stored in a text field, or even a simpler `DATE` field that should be parsed as `Y-m-d`

duncanmcclean:
Cool 👍 

One thing I've noticed is that if you don't have a `format` set on the date field in the blueprint, I seem to get a `Trailing data` error. As soon as I add one (`Y-m-d H:i:s`), it starts working. 

The field in question is the `created_at` column on my model. Could that be happening just because it's a special timestamp field? Maybe it doesn't happen if you have a normal `date` or `datetime` field, I've not tested with those two yet.

Another quick thing: do you think you'd be up for writing a quick test to make sure everything works as you'd expect? If not, that's fine, I can have a go.

Skullbock:
I should've fixed that, and added a test with 3 cases ;)

PR #61, opened 2021-07-28
Fix conflicting props / data name for columns

Skullbock:
## Description

Columns is a data name already use by the parent `Listing` vue component.
Let's rename the prop to `initialColumns` to avoid conflicts

### DIFF
--- a/resources/js/components/Listing/ListingView.vue
+++ b/resources/js/components/Listing/ListingView.vue
@@ -128,15 +128,15 @@ export default {
 
     props: {
         listingConfig: Array,
-        columns: Array,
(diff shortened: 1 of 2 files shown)


duncanmcclean (approved):
Thanks! Committed to resolve a conflict with #59.

Out of curiosity, was this causing issues for you or was it just a little thing you noticed while diving through the code? 😅

Skullbock:
This was causing issues to me on statamic 3.1.29 and i noticed the warning in the console log from Vue. Then i dove into it to discover what was generating it.

The issue appeared when trying to choose the columns to show on the index.

PR #62, opened 2021-07-30
don't json encode casted properties

Skullbock:
## Description

If in a model there is a casted property like array, collection, object, we don't need to json encode it, otherwise the caster will "double encode it"


duncanmcclean (approved):
Looks good, thanks!

PR #78, opened 2021-09-14
Null check on $record in toItemArray()

richwild:
## Description

When a relationship field contains a record which has been deleted, the entry publish form errors with `Call to a member function getKey() on null` at line 160.

This adds a null check on $record and returns the title as '[id] (not found)'.

## Related Issues

n/a

### DIFF
--- a/src/Fieldtypes/BaseFieldtype.php
+++ b/src/Fieldtypes/BaseFieldtype.php
@@ -155,11 +155,12 @@ class BaseFieldtype extends Relationship
         } else {
             $record = $id;
         }
-        
-        if (is_null($record)) {
+
+        if (! $record) {
             return [
                 'id' => $id,
-                'title' => $id . ' (not found)',
+                'title' => $id,
+                'invalid' => true,
             ];
         }
 
--- a/src/Fieldtypes/BaseFieldtype.php
+++ b/src/Fieldtypes/BaseFieldtype.php
@@ -155,6 +155,13 @@ class BaseFieldtype extends Relationship
         } else {
             $record = $id;
         }
+        
+        if (is_null($record)) {
+            return [
+                'id' => $id,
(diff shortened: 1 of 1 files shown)


src/Fieldtypes/BaseFieldtype.php
```diff
@@ -155,6 +155,13 @@ protected function toItemArray($id)
         } else {
             $record = $id;
         }
+        
+        if (is_null($record)) {
+            return [
```
duncanmcclean, line 160: Would you be able to return something like this instead? (as per the [Entry fieldtype](https://github.com/statamic/cms/blob/5b8d7724f45230598c4e7f010a2d32c0de309bbc/src/Fieldtypes/Relationship.php#L219))

```php
return [
    'id' => $id,
    'title' => $id,
    'invalid' => true,
];
```

I think passing in `invalid` means that the relationship field shows as red so the user can fix the relationship.

src/Fieldtypes/BaseFieldtype.php
```diff
@@ -155,6 +155,13 @@ protected function toItemArray($id)
         } else {
             $record = $id;
         }
+        
+        if (is_null($record)) {
```
duncanmcclean, line 159: Small thing: could you do `if (! $record)` instead?

duncanmcclean (changes requested):
Thanks for pointing this out! Just a couple things that need changing and I'll merge.

duncanmcclean:
I've went ahead and made the changes myself since they're tiny and I wanted to get it into the next release. Thanks again for this PR, I really appreciate it!

duncanmcclean:
Released & tagged in [v2.1.22](https://github.com/doublethreedigital/runway/releases/tag/v2.1.22)

PR #86, opened 2021-09-25
GraphQL

### LINKED ISSUE #54: Support for GraphQL

For headless sites with lots of entries, this could be the killer feature thats currently missing in Statamic. Being able to store some Collections in a database would be the perfect solution.

Would be great to be able to enable GraphQL for certain models. Sometimes we just add models that just hold data and shouldn't be publicly queryable.

Not that I can think of, is the data available in the content api? 

I'm going to need it at some point in the future but it's not major right now.

duncanmcclean:
## Description

This pull request introduces the ability to query your Runway resources/models with GraphQL.

## Enabling

GraphQL won't be enabled on resources by default, it'll have to be explicitly enabled with a simple config setting:

```php
// config/runway.php

'resources' => [
  \App\Models\Product::class => [
    'name' => 'Products',
    'blueprint' => 'product',

    'graphql' => true,
  ],
],
```

## Example Query

```graphql
{
  products {
    data {
      id
      name
      slug
      price
    }
  },
  product(id: 4) {
    name
    price
  }
}
```

## To Do

* [X] Base GQL Queries
* [X] Base GQL Type
* [x] Better way of deal with IDs and other non-blueprint columns
* [x] Figure out why queries are not showing in GraphiQL Explorer
* [x] Write a new 'GraphQL' page in the documentation (add to Oxygen once merged)
* [x] Support for the BelongsTo fieldtype
* [x] Support for the HasMany fieldtype
* [x] Add support for filtering
* [x] Add support for sorting

## Related Issues

Closes #54

### DIFF
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -116,11 +116,6 @@ class ServiceProvider extends AddonServiceProvider
                     return new \DoubleThreeDigital\Runway\GraphQL\ResourceShowQuery($resource);
                 });
 
-                // $this->app->bind("runway.graphql.types.{$resource->handle()}", function () use ($resource) {
-                //     return new \DoubleThreeDigital\Runway\GraphQL\ResourceType($resource);
-                // });
-
-                // GraphQL::addType("runway.graphql.types.{$resource->handle()}");
                 GraphQL::addQuery("runway.graphql.queries.{$resource->handle()}.index");
                 GraphQL::addQuery("runway.graphql.queries.{$resource->handle()}.show");
             });
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -113,7 +113,6 @@ class ServiceProvider extends AddonServiceProvider
                     return new \DoubleThreeDigital\Runway\GraphQL\ResourceType($resource);
                 });
 
-                // TODO: Add docblocks to this facade in core
                 GraphQL::addType("runway.graphql.types.{$resource->handle()}");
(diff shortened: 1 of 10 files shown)


src/ServiceProvider.php
```diff
@@ -92,4 +94,28 @@ protected function registerPermissions()
...
+
+                $this->app->bind("runway.graphql.queries.{$resource->handle()}.show", function () use ($resource) {
+                    return new \DoubleThreeDigital\Runway\GraphQL\ResourceShowQuery($resource);
+                });
+
+                $this->app->bind("runway.graphql.types.{$resource->handle()}", function () use ($resource) {
+                    return new \DoubleThreeDigital\Runway\GraphQL\ResourceType($resource);
+                });
+
+                GraphQL::addType("runway.graphql.types.{$resource->handle()}");
```
duncanmcclean, line 116: Maybe these need to happen before Statamic is booted in order for them to show in GraphiQL?

duncanmcclean:
Hmm, can't seem to get Runway's relationship fieldtypes to work in GraphQL, here's an example of a query I'm doing on the BelongsTo fieldtype:

```graphql
{
  product(id: 4) {
    id
    name
    brand_id
  }
}
```

And this is the response I get back from GraphQL:

```graphql
{
  "errors": [
    {
      "message": "Field \"brand_id\" of type \"runway.graphql.types.brand\" must have a sub selection.",
      "extensions": {
        "category": "graphql"
      },
      "locations": [
        {
          "line": 5,
          "column": 5
        }
      ]
    }
  ]
}
```

PR #92, opened 2021-10-20
Allow models w/ `has_many` relation to save.

### LINKED ISSUE #91: Can't save model that `has_many` relationship field
### Description

If you have a model with a `has_many` field, when you save, there is an error about at the `parent_field_id` can't be null.

### Steps to reproduce

1. Have a model w/ a `has_many` relation
2. Update model blueprint to include `has_many` field
3. Visit model in CP and save

### Environment

```
Statamic 3.2.16 Pro
Laravel 8.65.0
PHP 8.0.11
aryehraber/statamic-captcha 1.5.2
aryehraber/statamic-uuid 2.0.3
doublethreedigital/runway 2.1.x-dev
edalzell/download dev-main
edalzell/forma 1.0
jonassiewertsen/statamic-livewire 2.7.0
statamic/seo-pro 2.3.2
transformstudios/events 2.2.2
transformstudios/front 1.7.1
transformstudios/gated dev-main
transformstudios/magic-link dev-main
transformstudios/memorial dev-feature/orphan-portal
transformstudios/price-of dev-feature/orphan-portal
transformstudios/review 3.1
transformstudios/simple dev-main
transformstudios/stripe dev-main
transformstudios/takeover-redirects dev-main
```

edalzell:
## Description

The relation foreign id was being set to `null` when it should be set to the model id
## Related Issues

Closes #91

### DIFF
--- a/src/Fieldtypes/HasManyFieldtype.php
+++ b/src/Fieldtypes/HasManyFieldtype.php
@@ -62,13 +62,15 @@ class HasManyFieldtype extends BaseFieldtype
         $relatedResource = Runway::findResource($this->config('resource'));
         $relatedField = $record->{$this->field()->handle()}();
 
+        // I don't understand this code, why are you updating the reference? The `id` won't change.
         $relatedField
-            ->each(function ($model) use ($relatedField) {
+            ->each(function ($model) use ($record, $relatedField) {
                 $model->update([
-                    $relatedField->getForeignKeyName() => null,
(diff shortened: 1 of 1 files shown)


duncanmcclean:
I think remember why I did it now... In case you remove one of the related models, it'll set all the related relations as `null` initially, then it'll re-add all the relations again. 

I know it's not the best way to handle it but it's the way I could think of at the time.

PR #97, opened 2021-10-30
Make migration filename singular

### LINKED ISSUE #96: `runway:generate-migrations` creates migration filenames ending with `_tables` not `_table`
### Bug Description

As title. Migration gets created, but when you actually migrate the table an error "Cannot declare class CreateOrdersTable, because the name is already in use" is generated because the filename is tables not table.

### Steps to reproduce

1. Add resource like 
```php
        \App\Models\Order::class => [
            'name' => 'Orders',
            'blueprint' => 'orders',
        ],
```
2. Run `php please runway:generate-migrations order`
3. See order migration correctly created, but not the ending of the filename.
4. Run the migration.
5. See the error `PHP Fatal error:  Cannot declare class CreateOrdersTable, because the name is already in use in /var/www/html/database/migrations/2021_10_30_201449_create_orders_tables.php on line 7`
6. Discover the reason for the mysterious error on SO: https://stackoverflow.com/a/67595347/1281101
7. Make PR to fix

### Environment

Statamic 3.2.19 Pro
Laravel 8.68.1
PHP 8.0.7
doublethreedigital/runway 2.1.31
spatie/statamic-responsive-images 2.8.5

johncarter-:
## Description

Self explanatory.

## Related Issues

Closes #96

### DIFF
--- a/src/Console/Commands/GenerateMigration.php
+++ b/src/Console/Commands/GenerateMigration.php
@@ -320,7 +320,7 @@ class GenerateMigration extends Command
             ->__toString();
 
         File::put(
-            $migrationPath = database_path().'/migrations/'.date('Y_m_d_His').'_create_'.$resource->databaseTable().'_tables.php',
+            $migrationPath = database_path().'/migrations/'.date('Y_m_d_His').'_create_'.$resource->databaseTable().'_table.php',
             $migrationContents
         );
 


duncanmcclean (approved):
Nice catch - thanks! Released in [v2.1.32](https://github.com/doublethreedigital/runway/releases/tag/v2.1.32)

PR #99, opened 2021-11-15
Fix Statamic search when using HasMany fieldtype

edalzell:
## Description

If you have a `has_many` field in your resource, the CP search throws an error about that field is not a valid column (which is true).

At the moment this PR only filters out the `has_many` fields from the blueprint so the search works. I made a `searchableFields` so that this can be cleaned up and expanded as not all fields should be searchable (like asset paths etc). Perhaps it should use the "search index" feature of Statamic so that we can include/exclude/write transformers so that our Models can be searchable? Or maybe use a package like this: https://github.com/nicolaslopezj/searchable ?

I didn't have time to write tests yet, sorry. Making this PR so I can pull it in via composer patch.

HMU on discord to discuss how you'd like me to handle

### DIFF
--- a/src/Http/Controllers/ResourceListingController.php
+++ b/src/Http/Controllers/ResourceListingController.php
@@ -24,7 +24,7 @@ class ResourceListingController extends CpController
             abort(403);
         }
 
-        $sortField     = $request->input('sort', $resource->primaryKey());
+        $sortField = $request->input('sort', $resource->primaryKey());
         $sortDirection = $request->input('order', 'ASC');
 
         $query = $resource->model()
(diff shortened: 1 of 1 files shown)


duncanmcclean:
Can you please provide a bit of context around what was broken and how to reproduce it? 

Otherwise I have no idea what this PR is doing.

edalzell:
> Can you please provide a bit of context around what was broken and how to reproduce it?
> 
> Otherwise I have no idea what this PR is doing.

Add a `has_many` field to a Runway resource.
Do a search in the CP listing for that resource
See error

The error you see is because you grabbed all fields (which includes the relation) and tried a `like` comparison. But the relation field is not a column, so there's a SQL error about a invalid column (which is true).

This PR filters out that relation so that the comparison works on the other fields.

Happy to chat on discord if that helps.

duncanmcclean:
Thanks for the extra info - I can reproduce now!

Due to GH permissions I've had to move the branch into the Runway repo. I've opened a new pull request for it - you can subscribe for any updates/merges there: #100

PR #101, opened 2021-11-24
proper permission checks

edalzell:
## Description

The permissions for `store` and `destroy` are singular but the checks in the requests were plural, which caused an `Authorization failed` error.


duncanmcclean (approved):
Thanks!

PR #108, opened 2022-01-25
Feature: Allow toggling of create button in HasMany

ryanmitchell:
## Description

Allows you to determine whether the create button appears in HasMany field types.


duncanmcclean (approved):
Thanks!

PR #109, opened 2022-01-25
Feature: Allow HasMany to work on Stache entries

ryanmitchell:
## Description

Basically we by-pass all the logic you wrote and only run it if resourceHandle is available in the route, which it isnt on Stache entries.

This then allows the normal relationship field logic to run and the items get inserted as ids in the yaml array.

## Related Issues

None

### DIFF
--- a/src/Fieldtypes/HasManyFieldtype.php
+++ b/src/Fieldtypes/HasManyFieldtype.php
@@ -31,8 +31,9 @@ class HasManyFieldtype extends BaseFieldtype
     {
         $resourceHandle = request()->route('resourceHandle');
 
-        if (!$resourceHandle)
-            return $data;       
+        if (! $resourceHandle) {
+            return $data;      
+        }
         
         return collect($data)
             ->pluck('id')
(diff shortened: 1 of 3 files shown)


src/Fieldtypes/HasManyFieldtype.php
```diff
@@ -29,6 +29,11 @@ protected function configFieldItems(): array
     // Pre-process the data before it gets sent to the publish page
     public function preProcess($data)
     {
+        $resourceHandle = request()->route('resourceHandle');
```
duncanmcclean, line 32: I've not tested this yet but I'm wondering if this is going to work in all cases. 

For example: it might work for entries/terms but not sure if this check will work if it's a HasMany on another model? Eg. Post HasMany Tags..

ryanmitchell: Not sure I follow what you mean - `resourceHandle` should be available on any Runway defined resources/models - it was throwing an error when it wasn't defined (eg on entries/terms). Maybe I'm missing something...

duncanmcclean: Ignore me - it's probably me thinking up problems. I'll give it a proper test in the next few days. 

I appreciate the PR(s)!

duncanmcclean (approved):
Looks good - added a note to the documentation & added some code comments so I don't forget what's happening down the road.

Thanks again!

ryanmitchell:
Cool. So the concerns you had didn't apply? Just for my understanding...

duncanmcclean:
> Cool. So the concerns you had didn't apply? Just for my understanding...

Yeah, ignore my earlier comments. I don't know what I was worried about 😆

PR #111, opened 2022-01-31
Integrate with Statamic's CP Search

ryanmitchell:
## Description

This PR adds searching to the Statamic CP, assuming that the PR on the Statamic side of things gets merged. 

## Related Issues

Fixes: https://github.com/doublethreedigital/runway/issues/22

Corresponding PR on Statamic: https://github.com/statamic/cms/pull/5179


duncanmcclean (commented):
Hey,

Just got round to testing this PR but I can't seem to get any results to come from my Runway models. I'm testing with a resource that _does_ have routing enabled.

Is there anything I need to configure? I've switched my branches on both the Runway side and the CMS side.

ryanmitchell:
Did you add anything from runway to your default search config in Statamic and then rebuild the index?

e.g. mine is

    'indexes' => [

        'default' => [
            'driver' => 'local',
            'searchables' => 'runway:*',
            'fields' => ['title'],
        ],

    ],

once you do that and enable routing it should give results?

the registerSearchables() method in the runway SP needs updated to consider the fields requested in the $config var. At the  moment its mapping $result->brand to title which was my demo data.

the SearchResult data in ResourceRoutingRepository also needs updated to consider the fields. At the moment i have it returning dummy data.

I assumed you’d want to rearrange things so didnt do these bits… just wanted to get the proof of concept in place.

duncanmcclean:
@ryanmitchell 

No, I didn't have that. I've added it however and it's still not returning any results.

I've done a little bit of digging and doesn't seem like the custom Searchable in the Service Provider is being registered (`dd`s inside it aren't working). However, the right service provider is being used as syntax errors fail as they should.

I'll continue digging to see if I can figure it out..

ryanmitchell:
Let me know how you get on. I can share my repo if that helps.

duncanmcclean:
@ryanmitchell 

Thanks Ryan, I've got it working now. I was testing with `'searchables' => 'all'` which doesn't seem to work but `'searchables' => 'runway:*'` works fine.

I'm going to see if I can make it configurable.

src/Routing/ResourceRoutingRepository.php
```diff
@@ -3,16 +3,44 @@
...
+        [$resourceHandle, $id] = explode('::', $id, 2);
+
+        $resource = Runway::findResource($resourceHandle);
+        $record = $resource->model()->firstWhere($resource->primaryKey(), $id);
+
+        if (!$record) {
+            return null;
+        }
+
+        return (new SearchResult)->data([
```
duncanmcclean, line 22: I'm a little unsure about this always returning a `SearchResult`. 

While there's no actual issue with it at the moment, it means we wouldn't be able to do a normal lookup like `Data::find('runway::product::1')` and get back the associated 'record'. Well, we could but it would give us a SearchResult, not the model.

src/ServiceProvider.php
```diff
@@ -45,27 +48,26 @@ public function boot()
...
 
             $this->registerPermissions();
+            $this->registerSearchables();
             $this->bootGraphQl();
 
-            if (Runway::usesRouting()) {
-                $this->app->get(\Statamic\Contracts\Data\DataRepository::class)
-                    ->setRepository('runway-resources', Routing\ResourceRoutingRepository::class);
-            }
+            $this->app->get(\Statamic\Contracts\Data\DataRepository::class)
```
duncanmcclean, line 69: Just for note: I've made the `ResourceRoutingRepository` load for every site, even if no resources have routing enabled.

Before/after we merge this, it'll probably end up being renamed to the `ResourceRepository` instead.

src/ServiceProvider.php
```diff
@@ -95,6 +97,48 @@ protected function registerPermissions()
         }
     }
 
+    protected function registerSearchables()
+    {
+        Searchables::register('runway', function ($resource, $config) {
```
duncanmcclean, line 102: Another note: I've refactored this code to index the fields specified in the search config & I believe I've also made transformers work (I haven't tested them yet).

I've also refactored it to loop via a `Collection` instead of 'foreaching' through resources.

ryanmitchell:
Yep it’s not ideal but it was the way to make it work without big changes in statamic. 

I imagine Jason will change the approach or suggest a new method name anyway.

ryanmitchell:
Yep it's all looking good.

duncanmcclean (approved):
Thanks Ryan! This all looks good to me.

I've done a little bit of refactoring but essentially everything works exactly the same way. 

I think all we have to do now is to wait for the Statamic PR to be reviewed & merged in. We can keep this PR as a draft until that's done. 

Thanks again for all your work - I'm starting to understand more about the search stuff works under the hood by digging in 😆

duncanmcclean:
Once the Statamic implementation (https://github.com/statamic/cms/pull/5179) has been nailed, we'll need to come back around to this and make the various changes.

duncanmcclean:
I'm going to close this now as there's been some substantial changes on how the searchables stuff works in Statamic (statamic/cms#6318).

I believe @jasonvarga may open up a Runway PR once he's finished on the Statamic side 👀 

Thanks again though, @ryanmitchell!!

PR #119, opened 2022-02-16
Fix permission issues on the Listing Table and in the CP Nav

jbfournot:
# Hello everyone 👋🏻

This pull request fixes two issues and add little features :

## 1 - Better sidebar for resources based with policies
All resources are displayed in the sidebar including those whose the "view" policy is not allowed. I create inside the [Runway](src/Runway.php) main class a new method called `allResourcesAuthorized()` which return all resources authorized to be seen by a user. I take advantage of this correction to improve the display.

| Before | After |
| -------------- | :----------: |
|Screen Shot 2022-02-16 at 1 55 25 PM | Screen Shot 2022-02-16 at 1 51 42 PM |

In this example, the user is not allowed to view the discount resource. With the new `allResourcesAuthorized` function, this resources is not returned. Additionally, the design is the same as that of the collections section.

## 2 - Policies not applied in the ListingView component

I fixed the dropdown-list items inside the [ListingView](resources/js/components/Listing/ListingView.vue) component. The edit item did not reflect permissions. I take advantage of this correction to improve the display, for example, hide the dropdown menu when the user can't edit, delete or do an action (the menu was empty).  

Screen Shot 2022-02-16 at 1 51 52 PM


duncanmcclean (changes requested):
Hey - thanks for this!

However, I'd rather if we didn't change how Resources are displayed in the navigation. Users are used to how it looks at the moment and don't want to change it mid-version. 

Would you be able to remove that from this PR, then I'll give this a proper review?

duncanmcclean:
@jbfournot Just so you know, I've just pushed out a release with some options to configure Runway's CP Nav Items. [v2.3.3](https://github.com/doublethreedigital/runway/releases/tag/v2.3.3)

PR #124, opened 2022-02-26
Automatically set relation when creating with BelongsTo fieldtype

### LINKED ISSUE #112: Set belongsTo field automatically when creating

We have `Updates` that `hasMany`Attachments:
CleanShot 2022-02-02 at 16 00 16@2x

So the `Attachment` model `belongsTo` the Update:
```php
    public function orphanUpdate()
    {
        return $this->belongsTo(OrphanUpdate::class);
    }
```

When we Create & Link Item in the Update, I'd like the `orphan_update_id` on the `attachment` to be automatically set.

Happy to PR, with some guidance on how to start.

Not really, might consider moving the attachments into a json field on the update, but that'll likely not work either.

Blocker

duncanmcclean:
From now on, whenever you create a new related record using the 'Create & Link Item' button on the BelongsTo fieldtype, the relation will automatically be set on that new record.

Closes #112.

## Video

https://www.loom.com/share/757e6330b1274ff8a492169048dbcd0e

### DIFF
--- a/resources/js/components/Publish/PublishForm.vue
+++ b/resources/js/components/Publish/PublishForm.vue
@@ -112,53 +112,30 @@ export default {
 
     shouldShowSidebar() {
       return this.enableSidebar;
-
       // return this.enableSidebar && this.containerWidth > 920
     },
   },
 
   created() {
-    console.log("mount", this.publishContainer);
-
     // If we're creating a resource through the 'Create' on a HasMany field somewhere, fill any fields...
     if (this.publishContainer.includes("relate-fieldtype-inline")) {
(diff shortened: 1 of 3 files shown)


AndreasSchantl:
@duncanmcclean It would be really cool if this was the case with the HasMany field as well!

duncanmcclean:
> @duncanmcclean It would be really cool if this was the case with the HasMany field as well!

I thought that would have already been the case. Feel free to [open a feature request](https://github.com/duncanmcclean/runway/discussions/new?category=feature-requests) if it's not.

PR #132, opened 2022-03-17
Add missing bracket in documented example

brandon-9:
Simply fixed a missing array bracket in the documented example. Undocumented and using provided example breaks the site otherwise.


duncanmcclean:
Thanks!

PR #135, opened 2022-04-08
Store models in blink to prevent multiple DB calls

ryanmitchell:
This is a performance related PR. We are using runway on a site that often outputs the same items on a page. Instead of getting them each time, lets `Blink` them so we only do one database query per model on the front end.


duncanmcclean:
### Before

image

### After

image

Note: Brand is the relationship here

duncanmcclean (approved):
Thanks Ryan - will tag this shortly!

PR #142, opened 2022-04-26
Add limitRunwayQuery model to allow resources to be filtered/limited

ryanmitchell:
This PR adds support for a method to the index and field type listings that limits the results returned based on the developers requirements.

Our use case is we have a multi-site environment so we only want to return resources relevant to the site the user is editing. By providing this method we can filter the results.

## To Do

* [x] Made changes to the BaseFieldType and ResourceListingController files
* [ ] Added a test
* [ ] Updated the documentation

Haven't added tests or docs yet as I want to see if this is something you'd want to support.

### DIFF
--- a/src/Fieldtypes/BaseFieldtype.php
+++ b/src/Fieldtypes/BaseFieldtype.php
@@ -61,9 +61,16 @@ class BaseFieldtype extends Relationship
     {
         $resource = Runway::findResource($this->config('resource'));
 
-        return $resource->model()
-            ->orderBy($resource->primaryKey(), 'ASC')
-            ->get()
+        $model = $resource->model();
+        
+        $query = $model
+            ->orderBy($resource->primaryKey(), 'ASC');
+            
+        if (method_exists($model, 'limitRunwayQuery')) {
+            $model->limitRunwayQuery($query);
+        }
+                   
+        return $query->get()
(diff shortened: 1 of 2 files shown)


duncanmcclean:
Would you be able to add a code example of how this would look inside a model?

ryanmitchell:
Yeah of course... 

```
    public function limitRunwayQuery($query)
    {
        return $query->whereHas('categories')->where('something_else', true);
    }
```

ryanmitchell:
Yeah that makes sense. I've called the scope `runwayListing` to follow the same convention as `runwaySearch`.

```
    public function scopeRunwayListing($query)
    {
        return $query->whereHas('categories')->where('something_else', true);
    }
```

duncanmcclean:
Thanks - if you're able to, would you be able to add a test & mention it in the docs? I'll take a look tonight and get it merged in.

duncanmcclean (approved):
Thanks for this!

I ended up needing a couple of fixes in Runway after I pulled it into a project so I decided to merge this PR in now instead.

I spent some time trying to write some working tests but I couldn't figure it out because we needed to override the existing model just to add the scope.. I may loop back to it one day but should be fine without tests for now.

I've pulled in my changes from `2.3` and updated some of the wording in the docs.

PR #150, opened 2022-06-13
Correct behaviour of `runwaySearch` scope when searching in fieldtype

### LINKED ISSUE #149: scopeRunwaySearch doesnt override default behaviour
### Bug Description

When specifying a `scopeRunwaySearch` on my model I would expect it to override the default search behaviour that Runway provides. However, it seems that it provides[ search in **addition** to the default behaviour](https://github.com/doublethreedigital/runway/blob/7b9fcb97e677b5d29b7c4e04448fe548cb15fecd/src/Fieldtypes/BaseFieldtype.php#L90-L92).

This causes issues for our use case as it stops results being shown due to the subsequent `orWhere` clauses. 

There needs to be a way for the orWhere clauses not to be run

### Steps to reproduce

Should hopefully be sufficient detail in the description above.

### Environment

Statamic 3.3.12 Pro
Laravel 8.83.13
PHP 8.0.17
aerni/social-links 2.3.0
doublethreedigital/runway 2.3.5
jonassiewertsen/statamic-livewire 2.9.0
pecotamic/sitemap 1.4.0
statamic/eloquent-driver 0.2.0
statamic/seo-pro 3.1.0
thoughtco/statamic-minify 1.1.1
thoughtco/statamic-redirects 1.1.0

duncanmcclean:
This pull request fixes search in Runway's fieldtypes, when a `runwaySearch` scope is present on the resource's model. The behaviour is now the same as it is on the index listing table.

Now, if a `runwaySearch` scope exists, it'll be used, otherwise, it'll fallback to Runway's default search logic.

Fixes #149

### DIFF
--- a/src/Fieldtypes/BaseFieldtype.php
+++ b/src/Fieldtypes/BaseFieldtype.php
@@ -87,16 +87,20 @@ class BaseFieldtype extends Relationship
             ->when($request->search, function ($query) use ($request, $resource) {
                 $searchQuery = $request->search;
 
-                if ($query->hasNamedScope('runwaySearch')) {
-                    $query->runwaySearch($searchQuery);
-                }
-
-                $resource->blueprint()->fields()->items()->reject(function (array $field) {
-                    return $field['field']['type'] === 'has_many'
-                        || $field['field']['type'] === 'hidden';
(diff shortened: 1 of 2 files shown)


ryanmitchell:
Thank you :)

PR #154, opened 2022-06-29
Allow resources to have a title format

### LINKED ISSUE #153: Choose columns to display on `belongs_to` relationship

Right now we have a `belongs_to` relationship `orphan_id`:

```yaml
      -
        handle: orphan_id
        field:
          max_items: 1
          mode: select
          resource: orphan
          display: Orphan
          type: belongs_to
          icon: belongs_to
          listable: true
```

When displayed in the CP, it's only the `first_name` field:

CleanShot 2022-06-28 at 10 15 45@2x

As you can see, how can tell them apart if they have the same first name?

Don't think so, I tried adding `listable: true` to the `orphan` blueprint, but that didn't affect this field.

Will need it in the next few weeks

edalzell:
Closes #153 by adding a `title_format` to the `belongs_to` fieldtype so that you can set the title in the `select` mode of this fieldtype.

The format is Antlers.

## To Do

* [x] Made X change
* [ ] Added a test
* [ ] Updated the documentation

### DIFF
--- a/src/Fieldtypes/BaseFieldtype.php
+++ b/src/Fieldtypes/BaseFieldtype.php
@@ -261,7 +261,7 @@ class BaseFieldtype extends Relationship
 
     protected function makeTitle($record, $resource, string $default): string
     {
-        if (! $titleFormat = $resource->titleFormat()) {
+        if (! $titleFormat = $this->config('title_format')) {
             return $record->$default;
         }
 
--- a/src/Fieldtypes/BaseFieldtype.php
+++ b/src/Fieldtypes/BaseFieldtype.php
@@ -7,6 +7,7 @@ use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Arr;
 use Statamic\CP\Column;
 use Statamic\Facades\Blink;
+use Statamic\Facades\Parse;
 use Statamic\Fieldtypes\Relationship;
 
 class BaseFieldtype extends Relationship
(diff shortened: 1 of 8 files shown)


src/Fieldtypes/BelongsToFieldtype.php
```diff
@@ -17,6 +17,11 @@ protected function configFieldItems(): array
                 'default' => 1,
                 'read_only' => true,
             ],
+            'title_format' => [
```
duncanmcclean, line 20: Would you also be able to add this to the BelongsTo fieldtype?

src/Fieldtypes/BaseFieldtype.php
```diff
@@ -259,4 +258,13 @@ protected function getCreatables()
             ],
         ];
     }
+
+    protected function makeTitle($record, $resource, string $default): string
+    {
+        if (! $titleFormat = $this->config('title_format')) {
+            return $record->$default;
```
duncanmcclean, line 265: Instead of `$default` being passed into this method, could we instead simply use the first listable column?

duncanmcclean (changes requested):
Thanks for this! Suggested a couple of tweaks.

duncanmcclean (approved):
Thanks again for this! 

I added some tests and documentation for you, then realised I could just action the comments and have this tagged before you wake up, sorry for being impatient 😆 

Everything works the same way - just added tests, docs and added the config setting to the HasMany fieldtype as well.

edalzell:
Thanks, but I wasn’t done, found an issue yesterday. I’ll fix in another PR

duncanmcclean:
> Thanks, but I wasn’t done, found an issue yesterday. I’ll fix in another PR

Oh, sorry!

PR #181, opened 2022-11-04
Hide bulk delete button for read only resources

jymden:
This will pick out the first item from the items collection and use the single item visibility function to see if the delete button should be visible.  There might be more elegant ways to this but it seems to do the job and should also be able to handle null and non Model objects.


duncanmcclean (approved):
Thanks for this!

I've decided we should probably have a single source of truth for these checks so I've refactored the `visibleToBulk` method so it calls `visibleTo` for each item & if the count is the same at the end, the bulk actions are displayed.

PR #200, opened 2023-01-24
Search

jasonvarga:
This PR implements search using the features coming in Statamic 3.4.

I haven't updated the composer constraint in this PR but you'll need to .

**Searchable**

Statamic needs classes that implement `Searchable`. These would be classes like `Entry`, `User`, etc. In Runway's case, that would be the models, but since Runway doesn't actually touch the model classes, I've created a wrapper class.

**Provider**

The provider is what Statamic uses to find out what should go into the indexes.

The included class will take the appropriate resources, get all their models, and wrap them in those searchable classes.

When I say appropriate resources, I mean whatever you've configured to go in your search index:

```php
// config/statamic/search.php
'myindex' => [
  'driver' => 'local',
  'searchables' => ['collection:blog', 'runway:products'],
]
``` 

You can specify `runway:resource_handle` or `runway:*` to use all resources.

**Title field**

In order for search results to show in the CP, there needs to be some sort of title. But since your model might not have a `title` field, I've added a config option to pick which field is considered the title.

CleanShot 2023-01-24 at 12 53 09

```php
\App\Models\Order::class => [
  'name' => 'Orders',
  'title_field' => 'customer_name',
  // ...
]
```

This title is used in `Searchable@getCpSearchResultTitle`. You might want to change how that works completely, instead of using what I did.

**Augmentation**

The `search:results` tag loops over Result classes, which then does augmentation so it can let you use the blueprint fields etc. It assumes that the Searchable is augmentable.

You're kinda doing augmentation already in your tag. I've adjusted the AugmentedRecord class so it's actually considered an Augmentable. The searchable will then use that so it can be used in the search:results tag.

I bet we can simplify your tag and the AugmentedRecord class, but that can be done later.

**Updating indexes on save**

At the moment when you save a model in the CP, the search index doesn't get updated. I'm going to add something to the core to make that a bit simpler, and then I'll update the PR.

---

Let me know if you have any questions!

Todo:
- [x] The item index updating on save
- [x] The item index updating on delete


src/Search/Searchable.php
```diff
@@ -0,0 +1,80 @@
...
+    {
+        return vsprintf('runway::%s::%s', [
+            $this->resource->handle(),
+            $this->model->getKey(),
+        ]);
+    }
+
+    public function toSearchResult(): Result
+    {
+        return new ResultInstance($this, 'runway:'.$this->resource->handle());
```
jasonvarga, line 58: The second argument here is what you get when you use `{{ result_type }}` on the frontend.

Entries would be `entry`, terms would be `term`, etc.

Here I've used `runway:resourcehandle` so it wouldn't conflict if someone else was to have something called `resourcehandle`.

Maybe that's not important to you so you could remove the prefix, or maybe you want to change it to `runway_resourcehandle`.

duncanmcclean: Thanks! I'm good with the `runway:` prefix 👍

duncanmcclean (approved):

PR #214, opened 2023-02-01
feat: Make permission keys configurable

MtDalPizzol:
This PR tackles what was discussed [here](https://github.com/duncanmcclean/runway/discussions/211#discussion-4810850) 

- Add ability to override default permission keys
- Add ability to customize permission labels
- Add detailed docs on how to override permission keys, localize labels and add custom permissions
- Add a "build" script to package.json to allow building on newer Node.js versions

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,6 @@
 {
     "scripts": {
+        "build": "NODE_OPTIONS='--openssl-legacy-provider' npm run production",
         "dev": "npm run development",
         "development": "cross-env NODE_ENV=development node_modules/webpack/bin/webpack.js --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js",
         "watch": "npm run development -- --watch",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,5 @@
 {
     "scripts": {
-        "build": "NODE_OPTIONS='--openssl-legacy-provider' npm run production",
         "dev": "npm run development",
         "development": "cross-env NODE_ENV=development node_modules/webpack/bin/webpack.js --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js",
         "watch": "npm run development -- --watch",
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -67,6 +67,7 @@ class ServiceProvider extends AddonServiceProvider
             $this->registerPermissions();
             $this->registerNavigation();
             $this->bootGraphQl();
+
             SearchProvider::register();
             $this->bootModelEventListeners();
 
@@ -75,26 +76,6 @@ class ServiceProvider extends AddonServiceProvider
                     ->setRepository('runway-resources', Routing\ResourceRoutingRepository::class);
             }
         });
-    }
-
-    protected function permissionLabel($permission, $resource)
-    {
-        $translationKey = "runway.permissions.{$permission}";
-
-        $label = trans($translationKey, [
-            'resource' => $resource->name(),
-        ]);
-
-        if ($label == $translationKey) {
-            return match ($permission) {
-                'view' => "View {$resource->name()}",
-                'edit' => "Edit {$resource->name()}",
-                'create' => "Create {$resource->name()}",
-                'delete' => "Delete {$resource->name()}"
-            };
-        }
-
-        return $label;
     }
 
     protected function registerPermissions()
@@ -109,6 +90,7 @@ class ServiceProvider extends AddonServiceProvider
                             ->children([
                                 Permission::make("create {$resource->handle()}")
                                     ->label($this->permissionLabel('create', $resource)),
+
                                 Permission::make("delete {$resource->handle()}")
                                     ->label($this->permissionLabel('delete', $resource)),
                             ]),
@@ -162,5 +144,25 @@ class ServiceProvider extends AddonServiceProvider
                 Event::listen('eloquent.saved: ' . $class, fn ($model) => Search::updateWithinIndexes(new Searchable($model)));
                 Event::listen('eloquent.deleted: ' . $class, fn ($model) => Search::deleteFromIndexes(new Searchable($model)));
             });
+    }
+
+    protected function permissionLabel($permission, $resource)
+    {
+        $translationKey = "runway.permissions.{$permission}";
+
+        $label = trans($translationKey, [
+            'resource' => $resource->name(),
+        ]);
+
+        if ($label == $translationKey) {
+            return match ($permission) {
+                'view' => "View {$resource->name()}",
+                'edit' => "Edit {$resource->name()}",
+                'create' => "Create {$resource->name()}",
+                'delete' => "Delete {$resource->name()}"
+            };
+        }
+
+        return $label;
     }
 }
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -42,6 +42,10 @@ class ServiceProvider extends AddonServiceProvider
         Tags\RunwayTag::class,
     ];
 
+    protected $updateScripts = [
+        UpdateScripts\ChangePermissionNames::class,
+    ];
+
     public function boot()
     {
         parent::boot();
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -49,7 +49,7 @@ class ServiceProvider extends AddonServiceProvider
         $this->loadViewsFrom(__DIR__ . '/../resources/views', 'runway');
         $this->mergeConfigFrom(__DIR__ . '/../config/runway.php', 'runway');
 
-        if (!config('runway.disable_migrations')) {
+        if (! config('runway.disable_migrations')) {
             $this->loadMigrationsFrom(__DIR__ . '/../database/migrations');
         }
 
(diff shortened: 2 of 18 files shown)


package.json
```diff
@@ -1,5 +1,6 @@
 {
     "scripts": {
+        "build": "NODE_OPTIONS='--openssl-legacy-provider' npm run production",
```
duncanmcclean, line 3: Can we please remove this? It's not related to the change of how permissions work.

src/ServiceProvider.php
```diff
@@ -73,16 +73,56 @@ public function boot()
...
+    {
+        return match ($permission) {
+            'view' => "View {$resource->plural()}",
+            'edit' => "Edit {$resource->plural()}",
+            'create' => "Create new {$resource->singular()}",
+            'delete' => "Delete {$resource->singular()}"
+        };
+    }
+
+    protected function permissionKey($permission, $resource)
```
duncanmcclean, line 86: Out of interest, why do these permission keys need to be configurable, can't they just be fixed? Isn't it just the labels that need to be translated?

For example: it's just `view article`/`create article`, where `article` is the handle of the resource?

src/ServiceProvider.php
```diff
@@ -73,16 +73,56 @@ public function boot()
...
+    protected function permissionLabel($permission, $resource)
+    {
+        $translationKey = "runway.permissions.{$permission}";
+
+        $label = trans($translationKey, [
+            'resource' => $resource->name()
+        ]);
+
+        if ($label == $translationKey) {
+            return $this->legacyPermissionKey($permission, $resource);
```
duncanmcclean, line 104: I know you're trying to cleanup the code a little but can we move the code from the `legacyPermissionKey` method into this method?

duncanmcclean (changes requested):

package.json
MtDalPizzol: No problem. Would you like me to also fix the style checks that didn't pass? I'll have to learn how to configure PHP CS Fixer to conform to the repo style, but I can surely do this manually for now.

src/ServiceProvider.php
MtDalPizzol: I mean, it can be done. I'm just confused as why this would be beneficial...

MtDalPizzol: Oh... I see... Is it that you don't want a method call to something related to the key in the method that handles the label? In this case I think it makes sense.

src/ServiceProvider.php
duncanmcclean: In my example, I'm referring to the names of permissions, not the labels (which is what's seen in the Control Panel).

package.json
duncanmcclean: Yes, please!

src/ServiceProvider.php
MtDalPizzol: Yes... I understood that. And I also explained why the **names** must also be configurable, not only the **labels**.

`$user->hasPermission('View Autores')` or `$user->hasPermission('Create new Autore')` are examples of code that I have to write currently. `View Autores` and `Create new Autore` are the **NAMES** currently being generated by Runway, **NOT LABELS**.

**CURRENTLY:**
Name: 'Create new Autore'
Label: 'Create new Autore'

**WITH THIS PR WITHOUT CONFIGURING ANYTHING:**
Name: 'Create new Autore'
Label: 'Create new Autore'

**WITH THIS PR CONFIGURING **ONLY THE LABELS**:**
Name: 'Create new Autore'
Label: 'Criar autor'

**WITH THIS PR CONFIGURING **NAMES AND LABELS**:**
Name: 'create author'
Label: 'Criar autor'

duncanmcclean: Thanks, I think I understand now.

Instead of making those names configurable (like you have them now in the `runway.php` config file), can you hard code these names into Runway instead?

Statamic doesn't make them configurable for its permissions.

image

MtDalPizzol: Well... We could do that... But my thinking was that people already using the package had their permission names already stored in their stache files or their databases. Because of this, hardcoding this new permission names will cause their permissions to break when updating the package, because they will have permissions like "Create new Autore" saved on their stores, but the package will check for a "create author" permission.

That's why I chose the configurable path. Hardcoding this will have to come on a major release through a breaking change and, even then, maybe keeping this configurable would better.

duncanmcclean: Don't worry about that - I can add a migration script to change the permissions when folks update.

MtDalPizzol: IMO this is quite risky... But if you think this won't blow things up... Sure... We can try that...

duncanmcclean: It should be fine. It'll be included in a breaking release & will have a note in the changelog if people need to do it manually.

src/ServiceProvider.php
duncanmcclean: In fact, leave it how it is. Sorry!

MtDalPizzol:
Hey... @duncanmcclean I'm having trouble matching the code style in VSCode with PHP CS Fixer. Any guidance on that? This is creating a lot of unnecessary diffs that I don't want to commit.

duncanmcclean:
> Hey... @duncanmcclean I'm having trouble matching the code style in VSCode with PHP CS Fixer. Any guidance on that? This is creating a lot of unnecessary diffs that I don't want to commit.

I can fix it before merging if you'd like?

MtDalPizzol:
> > Hey... @duncanmcclean I'm having trouble matching the code style in VSCode with PHP CS Fixer. Any guidance on that? This is creating a lot of unnecessary diffs that I don't want to commit.
> 
> I can fix it before merging if you'd like?

Sure... but if you have any resource that can help me get this right... I'm planning more PRs after this one...

duncanmcclean:
Sometimes, if you click into StyleCI, it might give you the option for it to create a PR to your branch to fix any of the code style issues. I don't know if/how it works for forks. 

We're using the [Laravel preset](https://docs.styleci.io/presets#laravel) in StyleCI. I don't know if there's a way we can add a PHP CS Fixer file to the project which has the same rules on it - maybe that'll help?

Other than that - at the end of your PRs, you could work through StyleCI's suggested changes, do them in VS Code but save via the command palette (`File: save without formatting`).

Sorry it's a bit of a pain! 

PS: Sorry if I've been hard to work with on this PR - I think I was confused some of the time 😅

MtDalPizzol:
No problem. Runway is a package I'll heavily rely on for the biggest project my company ever took in. So, I'll be working fulltime on a project that will NEED to improve Runway together. This is why I'm being picky about conforming to the standards, so I don't have to waste time in the near future distracted by diffs caused by wrong formatting. I'm figuring it out here. Just asked for help because I spent a lot of time yesterday without getting anywhere with this. But things are coming together now. Hopefully everything will be good till the end of the day.

If you don't mind, I think adding a PHP CS Fixer file to the project's root is probably a good idea, since it can isolate the styles for the package from the styles for the app. But let's do this in another PR.

src/ServiceProvider.php
```diff
@@ -129,8 +141,28 @@ protected function bootModelEventListeners()
...
+
+    protected function permissionLabel($permission, $resource)
+    {
+        $translationKey = "runway.permissions.{$permission}";
+
+        $label = trans($translationKey, [
+            'resource' => $resource->name(),
+        ]);
+
+        if ($label == $translationKey) {
```
duncanmcclean, line 157: I've added an update script & documented the changes in the changelog.

Do we still need this `match` stuff here?

duncanmcclean (changes requested):
Nearly there... 

I've written an update script which will make the required changes to site's `roles.yaml` file. 

I've also made some small tweaks to the documentation around permissions & some of the code (nothing major!).

Before I merge this, I have one remaining question!

src/ServiceProvider.php
MtDalPizzol: I believe we do. If the user doesn't create a `resources/lang/{lang}/runway.php` file, this match is what will ensure that we show the default permissions labels in English. Otherwise, in the CP, we wouldn't even see the permission like name like `create resource`, but we would see `runway.permissions.create`, which is the key for the label in the language file.

duncanmcclean: Gotcha!

duncanmcclean (approved):
Thanks for this!

duncanmcclean:
On the StyleCI note - would you be able to let me know if [this PHP CS Fixer config](https://gist.github.com/mattias-persson/1be2f0537eab4ca5fb0ca0e788ec5a3d) works for you when making your next PR?

StyleCI doesn't provide a PHP CS Fixer equivalent of their presets as some of the rules aren't public.

MtDalPizzol:
> On the StyleCI note - would you be able to let me know if [this PHP CS Fixer config](https://gist.github.com/mattias-persson/1be2f0537eab4ca5fb0ca0e788ec5a3d) works for you when making your next PR?
> 
> StyleCI doesn't provide a PHP CS Fixer equivalent of their presets as some of the rules aren't public.

I think this is not ok. The main problem I had was the missing `not_operator_with_successor_space` rule which this config still doesn't apply.

What I did was to copy the StyleCI Laravel's preset rules and shove them into my PHP CS Fixer settings in VSCode. Still got some issues, like missing line break at file endings, but this was probably due to not using a base setting like `@PSR2`.

Also, I'm having trouble getting VSCode to use the config file from a specific project. It seems like the VSCode extension only allows you to specify a single CS Fixer config and doesn't check if the project root contains a config file. I ended up just configuring this globally.

So... Not sure if there's anything we can do to address this.

duncanmcclean:
Ah okay, that's a shame!

Usually when this happens to me, I make all my changes first. Then, at the end, I'll go through StyleCI's suggestions and fix them manually in my IDE (saving without formatting). 

Statamic also uses StyleCI and I think that's what they'd recommend too.

MtDalPizzol:
> Ah okay, that's a shame!
> 
> Usually when this happens to me, I make all my changes first. Then, at the end, I'll go through StyleCI's suggestions and fix them manually in my IDE (saving without formatting).
> 
> Statamic also uses StyleCI and I think that's what they'd recommend too.

No problem. I'll look into that to figure out whats the best workflow here... Thanks for taking the time to look into this anyway.

PR #216, opened 2023-02-04
Fix: Action Controller returning incorrect items to actions

Alt-Ben:
Tiny fix on getSelectedItems function to return the correct items, rather than first in DB  🙌


duncanmcclean (approved):
Thanks!

PR #218, opened 2023-02-05
Move authorization logic to a dedicated ResourcePolicy

MtDalPizzol:
This PR moves the authorization logic to a dedicated `ResourcePolicy`.

This will give devs the ability to customize authorization logic when needed.

Currently, if you need to customize how an action is authorized, you'll have to use the Service Container to rebind every class inside `src/Http/Request` plus every other class that makes checks using `->hasPermission()`.

Having a dedicated `ResourcePolicy` is also more aligned with Statamic, since the process to customize authorization will be exactly the same described in Statamic's docs to customize the `EntryPolicy`.

For instance:

```php
<?php
// app/Policies/ResourcePolicy.php

namespace App\Policies;

use DoubleThreeDigital\Runway\Policies\ResourcePolicy as RunwayResourcePolicy;
use Illuminate\Auth\Access\HandlesAuthorization;

class ResourcePolicy extends RunwayResourcePolicy
{
    use HandlesAuthorization;

    public function edit($user, $resource)
    {
       // Do custom checks here before running the default ones

        return parent::edit($user, $resource);
    }
}

```

```php
// app/Providers/AppServiceProvider.php

public function boot()
{
    $this->app->bind(RunwayResourcePolicy::class, ResourcePolicy::class);
}
```

PR #219, opened 2023-02-07
Prevent resource row selection

MtDalPizzol:
When clicking the edit link inside a row in a resource listing, before the editing page is loaded, the row gets briefly selected as if the user has clicked the row, not the link.

This will stop the click propagation from the link and prevent the selection from happening.


duncanmcclean (commented):
Thanks for this! I've never been able to figure out how to fix it before (maybe I just haven't tried hard enough! 🙈 )

It looks like your commits for your other pull request, #218 are showing up on this PR which means I can't merge this without also merging all those changes.

To workaround that, I've cherry-picked your commit and opened a PR for just that commit which I'll merge in shortly (https://github.com/duncanmcclean/runway/pull/220).

PR #221, opened 2023-02-08
Add dedicated ResourcePolicy

MtDalPizzol:
This PR moves the authorization logic to a dedicated `ResourcePolicy`.

This will give devs the ability to customize authorization logic when needed.

Currently, if you need to customize how an action is authorized, you'll have to use the Service Container to rebind every class inside `src/Http/Request` plus every other class that makes checks using `->hasPermission()`.

With this, you can simply extend and rebind the ResourcePolicy.

For instance:

```php
// app/Policies/ResourcePolicy.php

<?php

namespace App\Policies;

use DoubleThreeDigital\Runway\Policies\ResourcePolicy as RunwayResourcePolicy;
use Illuminate\Auth\Access\HandlesAuthorization;

class ResourcePolicy extends RunwayResourcePolicy
{
    use HandlesAuthorization;

    public function edit($user, $resource)
    {
        // Do custom checks here before running the default ones

        // If custom checks didn't return, run default checks
        return parent::edit($user, $resource);
    }
}
```

```php
// app/Providers/AppServiceProvider.php

<?php

namespace App\Providers;

use App\Policies\ResourcePolicy;
use DoubleThreeDigital\Runway\Policies\ResourcePolicy as RunwayResourcePolicy;

public function boot()
{
    $this->app->bind(RunwayResourcePolicy::class, ResourcePolicy::class);
}
```


duncanmcclean (approved):
Looks great, thanks for this!

PR #224, opened 2023-02-19
Exclude sections and computed values from search

stoffelio:
Stumbled across this today: when trying to search in a simple commerce resource (orders) I kept hitting SQL errors because section fields as well as computed fields were included and the app couldn't find columns for those in the database.


duncanmcclean (approved):
Thanks for this!

PR #255, opened 2023-04-21
Extra data option and access to the resource from scopes

royduin:
With this we can do things like:
```
AttributeOptionValue::class => [
    'name' => 'Manufacturers',
    'extra' => [
        'attribute_id' => 83,
    ],
],
```

```
public function scopeRunwayListing($query, $resource)
{
    return $query->where('attribute_id', $resource->extra()['attribute_id']);
}
```

~And with this creating an option to re-use models.~ https://github.com/duncanmcclean/runway/issues/256

### DIFF
--- a/docs/control-panel.md
+++ b/docs/control-panel.md
@@ -158,7 +158,7 @@ If you don't want to return everything, you may add a scope (`runwayListing`) to
 ```php
 class YourModel extends Model
 {
-	public function scopeRunwayListing($query)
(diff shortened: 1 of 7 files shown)


duncanmcclean:
Thanks for this!

However, I'm changing the way stuff is passed into `Resource`s in the next version of Runway, which is due to be released soon alongside Statamic 4 and I've decided to implement this a slightly different way, see: #257

royduin:
Awesome! Thanks!

PR #285, opened 2023-06-07
Fix bug in actions

ryanmitchell:
When using find()->first() it was getting the first resource by id, rather than the resource returned by find().

So in our delete action it was always deleting the first entry, not the one you actually had selected


ryanmitchell:
Ugh, ignore this, i'm obviously on an earlier branch!

duncanmcclean:
Haha 😂 

That was fixed a few weeks ago in #216.

PR #297, opened 2023-06-15
Allow for 'sort order' to be persisted on pivot tables

### LINKED ISSUE #287: Many to Many pivot table sort_order not updating
### Description

I am trying to build a LMS with Modules and Topics and want to create a Many to Many relationship where a Modules can have many Topics and a Topic can be assigned to many Modules.

I have 3 tables:
modules
topics
module_topic

When I add the hasMany field type to the Modules table I am able to create and link topics and the module_topics table is updated correctly.

However the sort_order column is not updated. 

I can update it in the DB manually and that will effect the order on the front end.

### Steps to reproduce

1. Create 2 Models
2. Add hasMany field to both
3. create pivot table in DB
4. Add one model to another
5. sort_order not updating in DB

### Environment

Environment
Application Name: Statamic
Laravel Version: 10.13.2
PHP Version: 8.2.4
Composer Version: 2.5.5
Environment: local
Debug Mode: ENABLED
URL: test.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 2
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.6.0 Solo

Statamic Addons
doublethreedigital/runway: 5.0.7
example/my-addon: dev-master

duncanmcclean:
In #259, you could configure a `order_column` which would be the column the sort order would be persisted to when saving a HasMany fieldtype.

However, this didn't cover the case where you might be using pivot tables with a Many to Many relationship. This pull request fixes that by implementing the re-orderable / sort column logic for relationships using pivot tables. 

Fixes #287.

### DIFF
--- a/docs/fieldtypes.md
+++ b/docs/fieldtypes.md
@@ -32,6 +32,7 @@ Also when configuring the fieldtype, you should choose the resource you wish to 
 
 -   **Eager Loading:** Using the `with` configuration option, you may specify any relationships you want to be eager loaded when the fieldtype is augmented.
 -   **Title Format:** Using the `title_format` configuration option, you may specify a title format to be used when viewing related results in the CP. You should use Antlers in this setting. (eg. `{{ first_name }} {{ last_name }}`)
+* **Reorderable** & **Order Column:** If you want to persist the ordering models, you may enable `reorderable` and specify the column used for storing the sort order (`order_column`).
 
 ### Table Mode
 
--- a/src/Fieldtypes/HasManyFieldtype.php
+++ b/src/Fieldtypes/HasManyFieldtype.php
@@ -125,6 +125,16 @@ class HasManyFieldtype extends BaseFieldtype
 
         // Many to many relation
         if ($relatedField instanceof BelongsToMany) {
+            // When Reordering is enabled, we need to change the format of the $data array. The key should
+            // be the foriegn key and the value should be pivot data (our sort order).
+            if ($this->config('reorderable') && $orderColumn = $this->config('order_column')) {
+                $data = collect($data)
+                    ->mapWithKeys(function ($relatedId, $index) use ($orderColumn) {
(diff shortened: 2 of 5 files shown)


tom-sk:
Thanks for updating this!

The sort order in the pivot table is now updating, but I'm having an issue displaying the order on the front end.

For it to update I have to set it to the name of the column in the pivot table:
` order_column: sort_order`

But for the order changes to appear on the front end it needs to be the name of the pivot table followed by the column name:
`order_column: module_topic.sort_order`

duncanmcclean:
> Thanks for updating this!
> 
> The sort order in the pivot table is now updating, but I'm having an issue displaying the order on the front end.
> 
> For it to update I have to set it to the name of the column in the pivot table: ` order_column: sort_order`
> 
> But for the order changes to appear on the front end it needs to be the name of the pivot table followed by the column name: `order_column: module_topic.sort_order`

Hmm, the ordering of models in the Control Panel is working fine for me when storing it on a pivot table 🤔 

For clarity, this is what my field looks like in my blueprint:

```php
[
                                'handle' => 'tags',
                                'field' => [
                                    'type' => 'has_many',
                                    'resource' => 'tag',
                                    'mode' => 'stack',
                                    'display' => 'Tags',
                                    'reorderable' => true,
                                    'order_column' => 'sort_order',
                                ],
                            ],
```

You don't happen to have a `sort_order` column on your `Module` or `Topic` models that Laravel might be trying to order with instead?

tom-sk:
Yeah that was it! I hade `sort_order` on both Module and Topics...

PR #312, opened 2023-08-04
[5.x] Support `search_index` tag supplemental data

ryanmitchell:
Fixes https://github.com/duncanmcclean/runway/issues/311 by supporting setSupplement and retrieval of supplemental data


duncanmcclean (approved):
Thanks for fixing this!

I've just added a `return $this` on the `supplement` method so `newAugmentedInstance` still works 😅

PR #324, opened 2023-09-10
[5.x] Fix validation issues

### LINKED ISSUE #321: Validation Issues: Nested Fields & DateTime Comparison
### Description

A couple of validation issues I've stumbled onto today:

1. Date comparisons seem to result in false-positives. That is, validation errors are thrown when they shouldn't be.
2. Within nested fields (like those found in a `Replicator` or `Grid`), input values don't seem to be validated against defined rules. 

### Steps to reproduce

**Date Comparisons**
1. Create two `Date` fields, `Start` and `End`, with time enabled.
2. Add a rule on the `Start` to make it first: `before:end`.
3. Try to save a record where `Start` _does_ come before `End`, but see validation error is thrown.

**Nested Fields**
1. Create a `Grid` field.
2. Within the `Grid`, create a `Text` field.
3. Within the field's validation rules, set the field to required.
4. Create a record, add a new row in the grid, but leave the text field blank.
5. Save and notice lack of validation errors.

[A potentially-relevant bit from Statamic's docs](https://statamic.dev/validation#validating-nestable-fields), if helpful.

### Environment

Environment
Application Name: TestingApp
Laravel Version: 10.21.0
PHP Version: 8.2.5
Composer Version: 2.5.8
Environment: local
Debug Mode: ENABLED
URL: testing-app.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Livewire
Livewire: v3.0.1

Statamic
Addons: 2
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.20.0 PRO

Statamic Addons
doublethreedigital/runway: 5.2.0
statamic/eloquent-driver: 2.6.1

Statamic Eloquent Driver
Asset Containers: file
Assets: file
Blueprints: file
Collection Trees: file
Collections: file
Entries: eloquent
Forms: file
Global Sets: file
Global Variables: file
Navigation Trees: file
Navigations: file
Revisions: eloquent
Taxonomies: file
Terms: file

duncanmcclean:
This pull request fixes a few validation issues pointed out in #321. A few examples of broken validation logic:

* A validation error wasn't being thrown when using the `before:date_field` validation rule. Now it is!
    * Although, I have noticed this validation rule seems to be thrown even when the first date field *is* before the other one. I can replicate on entries so it's a Core issue, not a Runway one.
* Validation errors weren't being thrown for fields inside Bards/Replicators/Grids, now they will be!

Closes #321.

### DIFF
--- a/src/Http/Controllers/ResourceController.php
+++ b/src/Http/Controllers/ResourceController.php
@@ -91,6 +91,13 @@ class ResourceController extends CpController
 
     public function store(StoreRequest $request, $resourceHandle)
     {
+        Runway::findResource($resourceHandle)
+            ->blueprint()
+            ->fields()
+            ->addValues($request->all())
+            ->validator()
+            ->validate();
+
         $postCreatedHooks = [];
 
         $resource = Runway::findResource($resourceHandle);
@@ -247,6 +254,13 @@ class ResourceController extends CpController
     public function update(UpdateRequest $request, $resourceHandle, $record)
     {
         $resource = Runway::findResource($resourceHandle);
+
+        Runway::findResource($resourceHandle)
+            ->blueprint()
+            ->fields()
+            ->addValues($request->all())
+            ->validator()
+            ->validate();
 
         $record = $resource->model()
             ->where($resource->model()->qualifyColumn($resource->routeKey()), $record)
--- a/src/Http/Requests/StoreRequest.php
+++ b/src/Http/Requests/StoreRequest.php
@@ -18,13 +18,4 @@ class StoreRequest extends FormRequest
 
         return User::current()->can('create', $resource);
     }
-
-    public function rules()
-    {
-        return Runway::findResource($this->resourceHandle)
-            ->blueprint()
-            ->fields()
-            ->validator()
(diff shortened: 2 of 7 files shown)


caseydwyer:
Thanks so much @duncanmcclean! 

> Although, I have noticed this validation rule seems to be thrown even when the first date field is before the other one. I can replicate on entries so it's a Core issue, not a Runway one.

This is a weird one...I `composer update`d everything, and that date-based before/after validation actually seems to be working correctly now, as far as I can tell. Even tested the time/minutes field and all seemed well. Maybe that just happened to get patched on Statamic's end too, in the last ~week or something? 🤔 Dunno, but working now—thanks again for the quick patch!

duncanmcclean:
Weird... maybe I was doing something wrong testing. Oh well, glad it's working!

PR #336, opened 2023-09-16
[5.x] Don’t save fields with `save` set to false

### LINKED ISSUE #331: Can't save model when there's a custom field
### Description

I have a custom field that is a button. There is no corresponding field in the DB.

When you try to save, you get a "missing field" DB error.

At a minimum, it should not try to save custom fieldtypes; however, there's an opportunity here to support custom fieldtypes.

First, in the blueprint there could be a field option `include` or something like that. And if it should be included, then you'd call a fieldtype method to get the data?

### Steps to reproduce

1. make a custom fieldtype
2. add to runway resource blueprint
3. try to save resource.

### Environment

Environment
Application Name: Zakat Foundation of America
Laravel Version: 10.23.0
PHP Version: 8.2.10
Composer Version: 2.5.5
Environment: local
Debug Mode: ENABLED
URL: zakat.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: daily
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 12
Antlers: runtime
Stache Watcher: Disabled
Static Caching: Disabled
Version: 4.21.0 PRO

Statamic Addons
aryehraber/statamic-captcha: 1.10.0
aryehraber/statamic-impersonator: 2.6.0
doublethreedigital/runway: 5.2.2
edalzell/forma: 2.1
jonassiewertsen/statamic-livewire: 2.12.0
statamic/seo-pro: 5.2.2
transformstudios/events: 5.0.1
transformstudios/front: 2.0.4
transformstudios/github: dev-main
transformstudios/neon: dev-feature/update-orphan-profile
transformstudios/review: 4.1.0
transformstudios/simple: 4.7.2

edalzell:
Closes #331 

Not sure where to put the new docs stuff, as the existing `fieldtypes.md` is all about the relationship stuff.

I can put it in there, or do you have another suggestion?

Note that the field we're using is `save` so that it's a positive and not negative one, but the code check to make sure it's actually defined so that it's backwards compatible.

### DIFF
--- a/src/Http/Controllers/ResourceController.php
+++ b/src/Http/Controllers/ResourceController.php
@@ -106,6 +106,10 @@ class ResourceController extends CpController
         foreach ($resource->blueprint()->fields()->all() as $fieldKey => $field) {
             $processedValue = $field->fieldtype()->process($request->get($fieldKey));
 
+            if (! $this->shouldSaveField($field)) {
+                continue;
+            }
+
             // Skip section fields or computed fields as there's nothing to store.
             if ($field->type() === 'section' || $field->visibility() === 'computed') {
                 continue;
@@ -269,6 +273,10 @@ class ResourceController extends CpController
         foreach ($resource->blueprint()->fields()->all() as $fieldKey => $field) {
             $processedValue = $field->fieldtype()->process($request->get($fieldKey));
 
+            if (! $this->shouldSaveField($field)) {
+                continue;
+            }
+
             // Skip section, HasMany and computed fields as there's nothing to store.
             if ($field->type() === 'section' || $field->type() === 'has_many' || $field->visibility() === 'computed') {
                 continue;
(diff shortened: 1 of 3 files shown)


edalzell:
Not sure why the tests failed, not related to this (cuz I didn't add any tests, yet)

duncanmcclean:
Out of interest, what's your use case for needing this over setting the button field's visibility to be `computed`?

duncanmcclean:
Thanks for the PR!

I've made a few tweaks but when adding a note to the documentation for this, it dawned on me that you can achieve the same thing using the native `visibility: computed` option on fields.

For most people, any fields without a real database column will be computed fields, where the data comes from an accessor. 

Although it doesn't completely match your use case, using the `visibility: computed` option would be the way I'd recommend ensuring Runway doesn't try and save the field's value to the database.

PR #353, opened 2023-10-20
[5.x] Don’t save fields with `save` set to false

edalzell:
Need to re-open #336.

New use case is that I need a field to be **not** saved **and** editable in the CP. Using `visibility:computed` means that the field is not editable in the CP.


edalzell:
Still need to update the tests as they reference `don't_save` which is not the correct config field handle

duncanmcclean:
> Still need to update the tests as they reference `don't_save` which is not the correct config field handle

No? `dont_save` is just the handle of the field in the blueprint we're using for testing. The setting itself is called `save`.

duncanmcclean (approved):

PR #355, opened 2023-10-27
[5.x] Use fields not items

edalzell:
If you use `items` in here you don't get the imported fields, if you use `fields()` you do.


duncanmcclean (approved):
Thanks for this!

I'll change everything else to use `->all()` during the v6 refactor - having the `Field` everywhere will be so much nicer than dealing with arrays. 

Anyways.... approve & merge ✅

PR #356, opened 2023-11-03
[6.x] REST API support

ryanmitchell:
This PR brings REST Api support to Runway.

Ensure the API is enabled, then opt into any resources you want in `statamic.api.resources` under a `runway` key:

```php
'resources' => [
    'collections' => true,
    'navs' => false,
    'taxonomies' => false,
    'assets' => false,
    'globals' => true,
    'forms' => false,
    'users' => false,
    'runway' => [
        'users' => ['allowed_filters' => ['name']],
    ],
],
```

Then your resource list should be available at `/api/runway/{resource}`

To do:

- [x] Tests 
- [x] Ensure model key is in the returned fields
- [x] Write documentation

### DIFF
--- a/src/Http/Controllers/RestApiController.php
+++ b/src/Http/Controllers/RestApiController.php
@@ -16,7 +16,7 @@ class RestApiController extends ApiController
 
     protected $resourceConfigKey = 'runway';
 
-    protected $routeResourceKey = 'resource';
+    protected $routeResourceKey = 'handle';
 
     protected $resourceHandle;
 
--- a/src/Http/Controllers/RestApiController.php
+++ b/src/Http/Controllers/RestApiController.php
@@ -1,0 +1,76 @@
+<?php
+
+namespace DoubleThreeDigital\Runway\Http\Controllers;
+
+use DoubleThreeDigital\Runway\Http\Resources\ApiResource;
+use DoubleThreeDigital\Runway\Resource;
+use DoubleThreeDigital\Runway\Runway;
+use Facades\Statamic\API\FilterAuthorizer;
+use Illuminate\Support\Str;
+use Statamic\Exceptions\NotFoundHttpException;
+use Statamic\Http\Controllers\API\ApiController;
+
+class RestApiController extends ApiController
+{
+    private $config;
+    protected $resourceConfigKey = 'runway';
+    protected $routeResourceKey = 'resource';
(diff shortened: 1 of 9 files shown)


src/Http/Controllers/RestApiController.php
```diff
@@ -0,0 +1,79 @@
...
+        return FilterAuthorizer::allowedForSubResources('api', $this->resourceConfigKey, Str::plural($this->resourceHandle));
+    }
+
+    private function makeResourceFromModel($resource, $model)
+    {
+        if (! $this->config) {
+            $this->config = collect(config('runway.resources'))->get(get_class($model));
+        }
+
+        return new Resource(
```
duncanmcclean, line 71: I'm a little confused as to why we're new-ing up a resource here, rather than using the ones discovered by the `Runway` class.

duncanmcclean (commented):
One thing I've noticed is that nested fields (those inside JSON columns) don't seem to work. Although, this may be unrelated to anything in this PR.

CleanShot 2023-11-06 at 20 13 54

Ideally, we'd have a `metadata` object, which inside it would have `alternative_title`, `publish_date`, `related_entries`.

src/Http/Controllers/RestApiController.php
ryanmitchell: Mainly just to be able to use the query builder stuff. Happy to do it differently if there's a better way.

ryanmitchell:
Not sure that's related to this PR. I'm using toAugmentedArray in the api resource so maybe it's something there?

src/Http/Controllers/RestApiController.php
duncanmcclean: Would [`Runway::findResourceByModel`](https://github.com/duncanmcclean/runway/blob/6.x/src/Runway.php#L84) do what you need?

duncanmcclean:
I'll try and carve out some time over the next week to give this a proper review - thanks for your work on this! 🙌 

I noticed that a BelongsTo field in my resource blueprint is causing an infinite loop when visiting the endpoint. That's probably something for me to figure out (may involve going back to shallow augmentation) 😅

duncanmcclean (approved):
Thanks for your work on this! 🙌

I've made a couple of nit-picky tweaks and fixed the issues I was having with nested fields & "BelongsTo" fields.

PR #367, opened 2023-11-07
[6.x] Only bind `resource` on cp routes

ryanmitchell:


duncanmcclean (changes requested):
Thanks for this PR 🙈 

Although it seems to be working fine in the Control Panel, the tests seem to be failing. Would you be able to take a look if you get a chance? No worries if not - I can loop back to it when I'm back on Runway.

ryanmitchell:
Yeah I will do - leave it with me.

duncanmcclean (approved):
Thanks!

PR #386, opened 2023-12-16
[5.x] Pass model to the policy

edalzell:
The `ResourcePolicy` needs the actual model, in case there's logic that requires it.


edalzell:
Needs some tests, I'll revisit next week, unless you wanna add some.

But this does work in my use case.

duncanmcclean:
> Needs some tests, I'll revisit next week, unless you wanna add some.

No worries! Let me know once the tests are in and I can review.

duncanmcclean (approved):

PR #393, opened 2024-01-08
[5.x] Bug: Missing hidden fields in listableColumns

### LINKED ISSUE #392: Missing "hidden" columns in listableColumns
### Description

Currently, if you mark a field as `hidden`, the second qualifier in the `listableColumns` method of `Resource.php` (noted below) causes it to be rejected. That is, it does not show up in the "column picker" of the control panel's index view:

```php 
/**
 * Returns an array of column handles that are marked as listable.
 */
public function listableColumns(): array
{
  return $this->blueprint()->fields()->items()
    ->reject(function ($field) {
      return isset($field['import'])
        // ❌ Remove this next line, to avoid rejecting hidden fields:
        || (isset($field['field']['listable']) && $field['field']['listable'] === 'hidden') 
        || (isset($field['field']['listable']) && $field['field']['listable'] === false);
      })
    ->pluck('handle')
    ->toArray();
}
```

PR in 2 minutes. 🙂

### Steps to reproduce

1. Create a model & blueprint with any field type.
2. Set `listable: hidden` on the field
3. Open the index page for the resource
4. Confirm missing/unavailable column in Customize Columns multi-picker

### Environment

<details>
<summary>Environment</summary>

Application Name: TF
Laravel Version: 10.39.0
PHP Version: 8.2.5
Composer Version: 2.5.4
Environment: local
Debug Mode: ENABLED
URL: tf.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Livewire
Livewire: v3.3.5

Statamic
Addons: 6
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.42.1 PRO

Statamic Addons
doublethreedigital/runway: 5.6.0
...

Statamic Eloquent Driver
Asset Containers: file
Assets: file
Blueprints: file
Collection Trees: file
Collections: file
Entries: eloquent
Forms: file
Global Sets: file
Global Variables: file
Navigation Trees: file
Navigations: file
Revisions: eloquent
Taxonomies: file
Terms: file
</details>

caseydwyer:
Fixes #392. Currently, if you mark a field as `hidden`, the second qualifier in the `listableColumns` method of `Resource.php` (noted below) causes it to be rejected. That is, it does not show up in the "column picker" of the control panel's index view:

```php 
/**
 * Returns an array of column handles that are marked as listable.
 */
public function listableColumns(): array
{
  return $this->blueprint()->fields()->items()
    ->reject(function ($field) {
      return isset($field['import'])
        // ❌ This PR removes the following line, to avoid rejecting hidden fields:
        || (isset($field['field']['listable']) && $field['field']['listable'] === 'hidden') 
        || (isset($field['field']['listable']) && $field['field']['listable'] === false);
      })
    ->pluck('handle')
    ->toArray();
}
```

This PR makes "column availability" on the index page behave like standard blueprints in Statamic. Here's [a snippet from the docs](https://statamic.dev/blueprints#unlisted-fields): 

> This will hide the field from entry listings by default, but still allows a user to toggle visibility using the column selector, and save those column preferences for his/her preferred workflow.

Let me know if I've missed anything here, or if this is intended behavior for any reason. Thanks Duncan!

Thanks!

### DIFF
--- a/src/Resource.php
+++ b/src/Resource.php
@@ -279,7 +279,6 @@ class Resource
         return $this->blueprint()->fields()->items()
             ->reject(function ($field) {
                 return isset($field['import'])
-                    || (isset($field['field']['listable']) && $field['field']['listable'] === 'hidden')
                     || (isset($field['field']['listable']) && $field['field']['listable'] === false);
             })
             ->pluck('handle')
--- a/src/Resource.php
+++ b/src/Resource.php
@@ -217,7 +217,19 @@ class Resource
     {
         return $this
             ->fluentlyGetOrSet('titleField')
-            ->getter(fn ($field) => $field ?? $this->listableColumns()[0])
+            ->getter(function ($field) {
+                if (! $field) {
+                    return collect($this->listableColumns())
+                        ->reject(function ($handle) {
+                            $field = $this->blueprint()->field($handle);
+
+                            return $field->get('listable') && $field->get('listable') === 'hidden';
+                        })
+                        ->first();
+                }
+
+                return $field;
+            })
             ->args(func_get_args());
     }
 
--- a/tests/ResourceTest.php
+++ b/tests/ResourceTest.php
@@ -123,4 +123,40 @@ class ResourceTest extends TestCase
 
         $this->assertSame($plural, 'Bibliotheken');
     }
+
+    /** @test */
+    public function can_get_listable_columns()
+    {
+        Config::set('runway.resources.DoubleThreeDigital\Runway\Tests\Post.blueprint.sections.main.fields', [
+            [
+                'handle' => 'values->normal_field',
+                'field' => ['type' => 'text'],
+            ],
+            [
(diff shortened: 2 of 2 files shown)


duncanmcclean (approved):
Thanks for this!

I've added some tests to cover this & added a listable check to the `titleField` method, which is used for figuring out which field should be the link in the listing table, which field should show in the fieldtypes, etc (it should always be a field that's shown by default).

caseydwyer:
> I've added some tests to cover this & added a listable check to the `titleField` method

Oh great call on the `titleField` bit—thanks for that, Duncan! 👍

PR #415, opened 2024-01-26
Fix add/remove markup in upgrade guide

vluijkx:
Noticed this small mistake in the docs while upgrading 🙂.


duncanmcclean (approved):
Whoops, thanks for fixing it!

Issue #412, opened 2024-01-24
MorphToMany relationships not displaying in front end.


mefenlon:
### Description

In runway v5, I was working with MorphToMany relationships and the front and back ends displayed correctly.
In v6, I am not unable to loop over the relationship on the front end despite the backend displaying correctly. 

Error in log:

> LOG.debug: Cannot loop over non-loopable variable: {{ titles }}

Displays fine on backend
<img width="746" alt="Screenshot 2024-01-24 at 2 57 41 PM" src="https://github.com/statamic-rad-pack/runway/assets/30707325/dd4da5c8-8053-4ae5-91f0-b02b9746228c">

Display on frontend
```
<div>
    Field name: {{ name }}
</div>
<div>
    Computed city_and_state: {{ city_and_state }}
</div>
<div>
    Title Relationship
    {{ titles }}
    {{ sort_title }}
    {{ /titles }}
{{#     LOG.debug: Cannot loop over non-loopable variable: {{ titles }} { #}}
{{#     "line": 9, #}}
{{#     "file": "\/var\/www\/html\/resources\/views\/runway\/city\/show.antlers.html" #}}
</div>
```
<img width="360" alt="Screenshot 2024-01-24 at 3 00 54 PM" src="https://github.com/statamic-rad-pack/runway/assets/30707325/139d547b-6e3a-4451-bad0-f1bbb1226e0a">

### Steps to reproduce

runway.php
```
<?php

return [

    /*
    |--------------------------------------------------------------------------
    | Resources
    |--------------------------------------------------------------------------
    |
    | Configure the resources (models) you'd like to be available in Runway.
    |
    */

    'resources' => [
         \App\Models\Title::class => [
             'cp_icon' => 'book-open',
             'name' => 'Titles',
             'route' => '/journals/{{ abbreviation }}',
             'title_field' => 'sort_title',
             'template' => 'runway.journal.show',
             'relationships' => ['cities'],
         ],
        \App\Models\City::class => [
            'cp_icon' => 'flag',
            'name' => 'Cities',
            'route' => '/cities/{{ slug }}',
            'title_field' => 'city_state_and_country',
            'template' => 'runway.city.show',
            'read_only' => true,
            'relationships' => ['titles'],
        ],
```
City Model
```
class City extends Model
{
    use HasRunwayResource;
    use RunwayRoutes;

    public $guarded = [];

    protected $appends = ['coordinates', 'city_state_and_country'];

    protected $with = ['titles'];

[...]

PR #416, opened 2024-01-31
[6.x] Fix model MorphToMany relationships not being augmented.

mefenlon:
This pull request fixes an issue where model MorphToMany relationships aren't being augmented after the augmentation #412


duncanmcclean (approved):
Thanks for this! 🙌 

Runway doesn't *really* support `MorphToMany` relationships. 

There's ~~probably~~ definitely other places where we don't have proper support for it (eg. no way to select models from across multiple resources in the fieldtype). Feel free to upvote #245.

However, I'll merge this PR for you so at least this part works and we're one step closer to possibly getting it fully working some day.

PR #419, opened 2024-02-01
[6.x] Fix model MorphMany relationships not being augmented. 

mefenlon:
Fixed augmentation for MorphMany relationships.

This got me thinking that `instanceof MorphTo` may also need to be accounted for, but I have not gotten to a use case on my code yet. I would assume it should in line 192 as

`if ($values instanceof BelongsTo  || $values instanceof MorphTo) `

I should encounter this eventually and will test then.


duncanmcclean (approved):
Thanks for this! We can add the `MorphTo` check when you end up needing it.

PR #425, opened 2024-02-06
[6.x] Fixed augmentation for BelongsToMany relationships

mefenlon:
Adds check for instance of BelongsToMany for augment and shallowAugment.


duncanmcclean (commented):
Thanks for this!

To prevent us from adding more and more conditions to these if statements, I've refactored it to check if `$value` is an instance of `Relation` instead, which all of the relationship classes do.

PR #427, opened 2024-02-10
[6.x] Adding support for query expressions in the 'order_by' config option

simonworkhouse:
This update provides support for providing a query expression to the `order_by` configuration option.

Here is an example of a potential use-case:

```php
'order_by' => (new \Illuminate\Database\Query\Expression('IFNULL(finalised_at, updated_at)')),
```

### DIFF
--- a/src/Resource.php
+++ b/src/Resource.php
@@ -2,6 +2,7 @@
 
 namespace StatamicRadPack\Runway;
 
+use Illuminate\Contracts\Database\Query\Expression;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\Schema;
@@ -106,7 +107,7 @@ class Resource
         return $this->config->get('read_only', false);
     }
 
-    public function orderBy(): string
+    public function orderBy(): string|Expression
     {
         return $this->config->get('order_by', $this->primaryKey());
     }


duncanmcclean:
Thanks for this PR!

I don't really *love* the ability to pass query expressions in Runway's config file, it doesn't feel like the right place for that sort of code. 

I haven't tested this but you can probably achieve this by adding a `runwayListing` query scope to your model, like this:

```php
public function scopeRunwayListing(Builder $query)
{
    $query->orderRaw('IFNULL(finalised_at, updated_at) DESC');
}
```

simonworkhouse:
@duncanmcclean Unfortunately that's not a workable solution as a sort order has already been applied prior to the `runwayListing` scope, defaulting to the primary key if a custom `order_by` has not been set. Technically speaking, it would be possible to replace the existing ordering at that point, but we wouldn't be able to tell whether or not a user was explicitly intending to order by the primary key field or not.

How about simply removing the `string` type definition, would that work for you? That would also keep it compatible with Laravel's `\Illuminate\Database\Query\Builder::orderBy($column, $direction = 'asc')` method which actually supports additional types with the following parameter definition:

```
@param \Closure|\Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder|\Illuminate\Contracts\Database\Query\Expression|string  $column
```

duncanmcclean:
Okay, I'd say that's a bug then. You should be able to do your own ordering in the `runwayListing` query scope, while have the user re-order things as they wish. 

The query probably needs tweaked to support that. I'd accept a PR to make that work.

PR #434, opened 2024-02-16
[6.x] Allow runwayListing scope to modify the order

ryanmitchell:
This PR moves the orderBy clause to after runwayListing has been run, and make it conditional on whether an orderBy clause already exists.

Let me know where to make a test for this as I cant see any coverage of the scopes?

Replaces https://github.com/statamic-rad-pack/runway/pull/433


simonworkhouse:
@ryanmitchell Unfortunately, this appears to have broken sort ordering in the admin panel for models that do not apply an order in the `runwayListing` scope. See https://github.com/statamic-rad-pack/runway/issues/437

PR #438, opened 2024-02-17
[6.x] Bring back handling user selection when there is no other orderby

### LINKED ISSUE #437: Resource listing sort order is not applied correctly
### Description

Sort ordering in the admin panel isn't being applied unless an order has already been applied by the `runwayListing` scope.

In the method `\StatamicRadPack\Runway\Http\Controllers\CP\ResourceListingController::index(...)` the following check only successfully applies a user selected sort order _if_ an order has already been applied to the query.

```php
        $query->when($query->getQuery()->orders, function ($query) use ($request) {
            if ($request->input('sort')) {
                $query->reorder($request->input('sort'), $request->input('order'));
            }
        }, fn ($query) => $query->orderBy($resource->orderBy(), $resource->orderByDirection()));
```

One can see that `$query->getQuery()->orders` will be `null` unless `$query->orderBy(...)` has already been called.

### Steps to reproduce

1. Attempt to apply a non-default sort order in the admin panel for a model that doesn't call `$query->orderBy(...)` in it's `runwayListing` scope.

See https://github.com/statamic-rad-pack/runway/commit/c9ec2aec506a195782f18c4924f84ce990266ecb for an examlpe test that reproduces this issue.

### Environment

Tests have been executed outside of an application environment.

ryanmitchell:
This [change](https://github.com/statamic-rad-pack/runway/pull/434/commits/0321d67eba5b65089926e1f52df56117aa49fa82) reverted some of the logic for handling user selected sorting.

This PR brings it back.

Closes #437

### DIFF
--- a/src/Http/Controllers/CP/ResourceListingController.php
+++ b/src/Http/Controllers/CP/ResourceListingController.php
@@ -30,7 +30,7 @@ class ResourceListingController extends CpController
             if ($request->input('sort')) {
                 $query->reorder($request->input('sort'), $request->input('order'));
             }
-        }, fn ($query) => $query->orderBy($resource->orderBy(), $resource->orderByDirection()));
+        }, fn ($query) => $query->orderBy($request->input('sort', $resource->orderBy()), $request->input('order', $resource->orderByDirection())));
 
         $activeFilterBadges = $this->queryFilters($query, $request->filters, [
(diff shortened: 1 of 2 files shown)


duncanmcclean (approved):
Whoops, my bad. Thanks for fixing it!

PR #443, opened 2024-02-24
[6.x] Fix permissions grouped under "Miscellaneous"

caseydwyer:
Prior to this PR, the permission set for Runway resources (eg, as seen when creating a new `role`) was grouped in with **Miscellaneous**, resulting in a duplicate header and no Runway section. This change updates the `registerPermissions` method in Runway's service provider, by swapping the trailing `->group` call for a `::group` closure/wrapper, as outlined in **[Statamic's docs](https://statamic.dev/extending/permissions#groups)**.

### Before:
image

### After:
image

### DIFF
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -114,23 +114,25 @@ class ServiceProvider extends AddonServiceProvider
 
     protected function registerPermissions(): self
     {
-        foreach (Runway::allResources() as $resource) {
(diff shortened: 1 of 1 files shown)


duncanmcclean (approved):

duncanmcclean:
Thanks! I'll include this in the next release.

caseydwyer:
Sounds good, thanks Duncan!

PR #448, opened 2024-02-29
[6.x] Performance issue with blueprints

### LINKED ISSUE #432: Listing models is slow
### Description

On our site we have ~1600 Orphans, and a large HasMany relationship too (`OrphanUpdates`).

When we try to list the orphans (locally on my M3 MacBookPro) it takes like 20 secs and I see this:
CleanShot 2024-02-15 at 16 19 28@2x

### Steps to reproduce

1. List Orphans
2. wait
3. wait
4. wait some more

### Environment

Environment
Application Name: Zakat Foundation of America
Laravel Version: 10.44.0
PHP Version: 8.3.2
Composer Version: 2.5.5
Environment: local
Debug Mode: ENABLED
URL: zakat.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: daily
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 11
Antlers: runtime
Sites: 3 (zakat.org (EN), zakat.org (AR), zfinstitute.com (ZFI))
Stache Watcher: Disabled
Static Caching: Disabled
Version: 4.48.0 PRO

Statamic Addons
aryehraber/statamic-captcha: 1.10.0
edalzell/forma: 2.1
jonassiewertsen/statamic-livewire: 2.12.0
statamic-rad-pack/runway: 6.1.0
statamic/seo-pro: 5.4.1
transformstudios/events: 5.3.0
transformstudios/front: 2.0.4
transformstudios/github: dev-main
transformstudios/neon: dev-fix/upgrade-runway
transformstudios/review: 4.1.4
transformstudios/simple: 4.20.1

edalzell:
When looping through each field, the blueprint was retrieved each time, even though it's static. This causes a performance issue because that method was called many times for each model.

Now it's retrieved once.

Closes #432.

### DIFF
--- a/src/Http/Resources/ResourceCollection.php
+++ b/src/Http/Resources/ResourceCollection.php
@@ -5,7 +5,6 @@ namespace StatamicRadPack\Runway\Http\Resources;
 use Illuminate\Http\Resources\Json\ResourceCollection as LaravelResourceCollection;
 use Statamic\Facades\Action;
 use Statamic\Facades\User;
-use Statamic\Fields\Blueprint;
 use StatamicRadPack\Runway\Fieldtypes\BelongsToFieldtype;
 use StatamicRadPack\Runway\Fieldtypes\HasManyFieldtype;
 use StatamicRadPack\Runway\Runway;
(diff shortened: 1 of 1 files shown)


duncanmcclean:
Thanks for this PR!

However, after comparing Runway's `ResourceCollection` to [the equivalent code](https://github.com/statamic/cms/blob/4.x/src/Http/Resources/CP/Entries/Entries.php) in Statamic Core, I've decided that Runway's implementation probably needs refactoring to be more like Core.

This refactoring should hopefully solve both of the main problems we discussed (the wrong columns being sent through to `preProcessIndex` & blueprints being loaded again for every model).

Replacement PR: #451

PR #449, opened 2024-02-29
[6.x] Prevent all listable columns from being processed, even when they're not needed

duncanmcclean:
This pull request fixes an issue where all listable columns were being processed in the `ResourceCollection@toArray` method.

Even though we did have a check to ensure columns existed in the `$columns` array before actually processing them, it turns out the `$columns` array actually included columns we don't need/want.

Instead, we're now checking against the `requestedColumns()` method (provided by the `HasRequestedColumns` trait) which will give us the columns provided in the request, then fallback to the columns from the blueprint.


duncanmcclean:
Replacing with #451, which includes the changes in this PR but refactors how everything else works.

PR #451, opened 2024-02-29
[6.x] Refactor how Control Panel listing data is loaded

duncanmcclean:
This pull request refactors how the data for the Control Panel listings work, bringing things more inline with how Statamic handles it for entries.

These changes should *hopefully* solve a lot of the performance problems #448 & #449 were attempting to sort. The blueprint is no longer being re-fetched on each model & we're now only preprocessing data for columns we're actually using.

Replaces #448.
Replaces #449.
Closes https://github.com/statamic-rad-pack/runway/issues/432.


edalzell:
This is SO FAST NOW!!!!!!!

PR #453, opened 2024-03-04
Update form requests for new route/model binding

edalzell:
V6 changed the route/model binding for the`statamic.cp.runway.update` and `statamic.cp.runway.edit` routes from `record` to model`.

`EditRequest` and `UpdateRequest` were not updated to match.

I didn't see any tests for the policies, but I can add if you let me know how you'd like those done.


edalzell:
I have confirmed this works on our site

PR #454, opened 2024-03-05
[6.x] Scope for generating runway route uri

mefenlon:
- Adds ability to scope resources for uri generation when running command `php please runway:rebuild-uris`
- Adds documentation for defining the scope in your model.

### DIFF
--- a/docs/frontend-routing.md
+++ b/docs/frontend-routing.md
@@ -28,6 +28,20 @@ class Product extends Model
     use RunwayRoutes;
 ```
 
+## Scoping Front-end Routing URI Generation
+
+If you don't want to generate a uri for everything, you may add a scope (`runwayRoutes`) to limit the returned results.
+
+```php
+class YourModel extends Model
+{
+	public function scopeRunwayRoutes($query)
+	{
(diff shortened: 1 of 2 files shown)


duncanmcclean:
Thanks for this - I'll try and find time to review this in the next few days.

While this limits the models returned in the `runway:rebuild-uris` command, it doesn't prevent rows from being [added to the `runway_uris` table on save](https://github.com/statamic-rad-pack/runway/blob/6.x/src/Routing/Traits/RunwayRoutes.php#L76). However, maybe that's not a big deal?

mefenlon:
The add on save is not an issue for my current application. Basically I am using a subest (one of a few project tags) of a larger database. Working with the full dataset, I am using a global Scope on the model. Perhaps expanding the functionality to have this act as a global (for runway) scope makes sense.

duncanmcclean (approved):
Thanks for this PR! 🙌 

I've tweaked the code slightly and have also added a whole section to the "Front-end Routing" page about the URI Cache since I realised we don't actually mention it anywhere.

PR #455, opened 2024-03-08
[6.x] Accept --no-interaction option for the runway:rebuild-uris command and disable user prompt

mefenlon:
Disables the confirm prompt in the runway:rebuild-uris command. Useful for scripting the command or programmatically executing command.

### DIFF
--- a/src/Console/Commands/RebuildUriCache.php
+++ b/src/Console/Commands/RebuildUriCache.php
@@ -45,12 +45,14 @@ class RebuildUriCache extends Command
      */
     public function handle()
     {
-        $confirm = $this->confirm(
-            'You are about to rebuild your entire URI cache. This may take part of your site down while running. Are you sure you want to continue?'
-        );
+        if (!$this->option('no-interaction')) {
+            $confirm = $this->confirm(
+                'You are about to rebuild your entire URI cache. This may take part of your site down while running. Are you sure you want to continue?'
+            );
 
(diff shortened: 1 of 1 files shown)


src/Console/Commands/RebuildUriCache.php
```diff
@@ -45,12 +45,14 @@ public function __construct()
      */
     public function handle()
     {
-        $confirm = $this->confirm(
-            'You are about to rebuild your entire URI cache. This may take part of your site down while running. Are you sure you want to continue?'
-        );
-
-        if (! $confirm) {
-            return;
+        if (!$this->option('no-interaction')) {
```
duncanmcclean, line 48: Can we use `--force` instead?

Laravel does this for its commands, like [`migrate:fresh`](https://github.com/laravel/framework/blob/10.x/src/Illuminate/Database/Console/Migrations/FreshCommand.php#L111) which need user confirmation before doing anything.

duncanmcclean (changes requested):

src/Console/Commands/RebuildUriCache.php
mefenlon: Absolutely. Just made the changes and added description to input options.

mefenlon (commented):
Fixed style and changed option name to force

duncanmcclean (approved):
Thanks!

PR #457, opened 2024-03-11
[6.x] GraphQL: add resolver for fields with `_id`

arcs-:
The docs for [Belongs To](http://runway.duncanmcclean.com/fieldtypes) mention that a reference to a Statamic Entry doesn't need to be in the format of `model_id`, this is not true as it won't resolve it at all with the `_id` suffix as it won't be appended when resolving it.

This change would attempt to resolve the property with the `_id` suffix, so it would be possible to have a column called `model_id`.

### DIFF
--- a/src/GraphQL/ResourceType.php
+++ b/src/GraphQL/ResourceType.php
@@ -46,6 +46,10 @@ class ResourceType extends Type
                 $model = $resource->model()->firstWhere($resource->primaryKey(), $model);
             }
 
+            if (isset($model->{$info->fieldName.'_id'})) {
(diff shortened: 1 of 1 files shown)


duncanmcclean:
Sorry, I'm not sure I'm completely following the issue. Would you be create a fresh Statamic site to demonstrate the issue so I can easily reproduce. Thanks!

duncanmcclean:
> Sorry, I'm not sure I'm completely following the issue. Would you be create a fresh Statamic site to demonstrate the issue so I can easily reproduce. Thanks!

I'm going to close this PR for now. When you're able to demonstrate the issue this fixes, I'm happy to re-open. Thanks!

PR #466, opened 2024-03-22
[6.x] Fix incorrect collection handle plural

Ken-vdE:
And added example that better relates to the rest of the page


duncanmcclean (approved):
Thanks!

PR #470, opened 2024-04-09
[6.x] Fix N+1 queries in ResourceActionController

### LINKED ISSUE #469: N+1 queries in ResourceActionController
### Description

Currently, in the `getSelectedItems` method in the `ResourceActionController`, we loop over selected IDs (ie, `$items`) and then `find` each corresponding record. This results in N+1 queries, and becomes slower/more problematic for larger selections. See debugbar's query counter in this example—over 50 queries are processed, when the whole page is selected:

n-plus-1-before

PR queued up to patch, hang tight.

### Steps to reproduce

1. Set up a resource with an action
2. Go the list view, and open debugbar (or monitor queries somehow)
3. Select the entire page, watch query-count jump way up

### Environment

<details>
  <summary>Environment</summary>

Environment
Application Name: TF
Laravel Version: 10.48.3
PHP Version: 8.2.5
Composer Version: 2.5.4
Environment: local
Debug Mode: ENABLED
URL: tf.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single, bugsnag
Mail: smtp
Queue: database
Session: database

Livewire
Livewire: v3.4.9

Statamic
Addons: 6
Antlers: runtime
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.53.2 PRO

Statamic Addons
doefom/currency-fieldtype: dev-main
eminos/statamic-group: 2.0.1
eminos/statamic-tabs: 1.1.0
jacksleight/statamic-miniset: 1.1.1
statamic-rad-pack/runway: 6.6.1
statamic/eloquent-driver: 2.10.0

Statamic Eloquent Driver
Asset Containers: file
Assets: file
Blueprints: file
Collection Trees: file
Collections: file
Entries: eloquent
Forms: file
Global Sets: file
Global Variables: eloquent
Navigation Trees: file
Navigations: file
Revisions: eloquent
Taxonomies: file
Terms: file
</details>

caseydwyer:
Closes #469. After this change, models are essentially eager-loaded, using the `findMany` method inside `getSelectedItems`. See #469 for a before, and here's an after (see query counter in the debugbar):

n-plus-1-after

This also has (what I would consider) an added benefit, by returning an _eloquent_ collection, instead of a generic collection. Allows you to do some handy things within the `Action` class itself, like `$items->load('anotherRelation')`. Wanted to mention this in case it impacts anything downstream that I'm unaware of, but should just be a net-positive since the eloquent collection class extends the base collection class. 👍

Let me know if we need to tweak anything, here!

### DIFF
--- a/src/Http/Controllers/CP/ResourceActionController.php
+++ b/src/Http/Controllers/CP/ResourceActionController.php
@@ -26,6 +26,6 @@ class ResourceActionController extends ActionController
 
     protected function getSelectedItems($items, $context)
     {
-        return $items->map(fn ($item) => $this->resource->find($item));
+        return $this->resource->findMany($items);
     }
 }


duncanmcclean:
Thanks for this! I'll review it when I can.

PR #496, opened 2024-05-16
[7.x] Check if field is read_only

mefenlon:
Checks if a HasManyFieldtype is set to read_only


duncanmcclean:
Thanks for this pull request!

However, I don't think this is necessarily the right fix.....

When fieldtypes are set to read-only, they should still be calling the fieldtype's `process` method, since technically changes could still be made (eg. with custom JavaScript), even if the user can't edit the field directly.

If you don't want changes to be saved, add `save: false` to your field's config.

PR #508, opened 2024-06-04
[7.x] Prevent models from being created by setting blueprint to be hidden

ryanmitchell:
~~This PR adds a `can_create` option similar to `read_only` but only affecting creation.~~

Sometimes you want to allow viewing and editing of items, but not creation of new ones.  As policies don't apply to super users this can't be satisfactorily handled through returning false on a policy create function.

This PR adds a check for whether the blueprint is hidden or not, and if so it hides the create button in the listing view and prevents new models from being created through the CP.


duncanmcclean:
Thanks for this pull request!

On one hand, this seems like a pretty small, low-risk PR.

However, on the other hand, it introduces a difference between how this situation is handled in Runway vs native entries, which is something I'm keen to avoid where possible. 🤔 

As you mentioned, permissions (& policies) would be the recommended way of handling this situation in native entry land, it just wouldn't work for super users since they can access everything. 

Maybe there's an opportunity for adding a similar option to collections/entries, then we can mirror it on the Runway side?

duncanmcclean:
Leaving it open in case I change my mind.

ryanmitchell:
> As you mentioned, permissions (& policies) would be the recommended way of handling this situation in native entry land, it just wouldn't work for super users since they can access everything.

Yep the ideal solution would be that the super user gate in core is reworked so that policies can override that behaviour and then the same approach could work in both core and Runway. But as that would be considered a breaking change it's at best a year away now.

Anyway, I've remembered there is a way in core of hiding the create button and preventing creation - by having no blueprints that are visible (ie they all have hide: true). So I've updated the logic here to check if the blueprint is hidden or not.

duncanmcclean:
> Anyway, I've remembered there is a way in core of hiding the create button and preventing creation - by having no blueprints that are visible (ie they all have hide: true). So I've updated the logic here to check if the blueprint is hidden or not.

Ah, so there is, well remembered!

duncanmcclean (approved):
Thank you! This works great. 🥳

PR #524, opened 2024-06-13
[7.x] Allow search_indexes on resources

ryanmitchell:
This PR adds the ability to specify a `search_index` on resources to use a Statamic search index in both the CP field type and the listing view.

Closes https://github.com/statamic-rad-pack/runway/discussions/501

### DIFF
--- a/docs/resources.md
+++ b/docs/resources.md
@@ -168,6 +168,19 @@ If you'd like to specify a different field, you may do so by setting the `title_
 ],
 ```
 
+### Search index
+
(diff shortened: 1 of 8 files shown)


duncanmcclean (changes requested):
Thanks again for this pull request!

I've added a test to cover the new `search_index` setting in both of the fieldtype's. I also added `Runway::discoverResources()` to your test, otherwise the `search_index` setting wouldn't actually end up being set on the resource. 

However, it seems like all three of these tests are failing. Would you be able to take a look when you get a chance?

duncanmcclean (approved):

PR #553, opened 2024-07-09
[7.x] Add Publish State permission

edalzell:
Matching the `canManagePublishState` from core and replicating this PR: https://github.com/statamic/cms/pull/10424

### DIFF
--- a/resources/js/components/Publish/PublishForm.vue
+++ b/resources/js/components/Publish/PublishForm.vue
@@ -103,7 +103,7 @@
                                     class="flex items-center justify-between px-4 py-2"
                                     :class="{ 'border-t dark:border-dark-900': resourceHasRoutes && permalink }"
                                 >
-                                    <label v-text="__('Published')" class="font-medium publish-field-label" />
+                                    <label v-text="__('Published')" class="publish-field-label font-medium" />
                                     <toggle-input :value="published" :read-only="!canManagePublishState" @input="setFieldValue(resource.published_column, $event)" />
                                 </div>
                             </div>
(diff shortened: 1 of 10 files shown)


edalzell:
@duncanmcclean let's talk about this one, as I believe we'd need a migration script otherwise it's breaking.

duncanmcclean:
> let's talk about this one, as I believe we'd need a migration script otherwise it's breaking.

Yes, we'd either need to ship with an upgrade script or include this PR in the next breaking changes release (which'll have the nested fields fix).

src/ServiceProvider.php
```diff
@@ -149,6 +149,9 @@ protected function registerPermissions(): self
                                     Permission::make("create {$resource->handle()}")
                                         ->label($this->permissionLabel('create', $resource)),
 
+                                    Permission::make("publish {$resource->handle()}")
```
duncanmcclean, line 152: The concept of "publish states" are opt-in, so we probably only want to add this permission if they're enabled for the resource.

```php
if ($resource->hasPublishStates()) {
    Permission::make("publish {$resource->handle()}")
        ->label("Manage {$resource->handle()} Publish State"),
}
```

resources/js/components/Publish/PublishForm.vue
```diff
@@ -158,8 +158,8 @@
                     class="rtl:mr-4 ltr:ml-4 btn-primary flex items-center"
                     :disabled="!canPublish"
                     @click="confirmingPublish = true">
-                    <span v-text="__('Publish')" />
-                    <svg-icon name="micro/chevron-down-xs" class="rtl:mr-2 ltr:ml-2 w-2" />
+                    <span v-text="this.canManagePublishState ? __('Publish') : __('Create Revision')" />
```
duncanmcclean, line 161: Like we've done in Core, can we extract this out into a `publishButtonText` method?

resources/js/components/Publish/PublishForm.vue
duncanmcclean, line 1: Related to [my comment here](https://github.com/statamic/cms/pull/10424#discussion_r1672091778), you updated the text of multiple buttons in this PR, but only updated one button in your Core PR.

Is there any particular reason for this?

duncanmcclean (changes requested):

duncanmcclean (approved):

PR #564, opened 2024-07-18
[7.x] Resource Actions

edalzell:
Related: https://github.com/statamic-rad-pack/runway/discussions/563
Related: statamic/cms#10471


resources/js/components/Listing/InlineActions.vue
duncanmcclean, line 1: I might have missed something, so apologies if I have, but is there any particular reason we need our own `InlineActions` component and we can't just use the one in Core?

duncanmcclean:
Thanks for this!

The PR looks good. However, I'm just going to wait out for statamic/cms#10471 to be merged, in case Jason makes any changes we should copy over to the Runway implementation.

duncanmcclean:
> Statamic PR has been merged!

Just saw that - will take a look at this PR this week.

duncanmcclean (approved):

PR #577, opened 2024-07-30
[7.x] Fix JSON field revisions

edalzell:
Due to https://github.com/statamic-rad-pack/runway/pull/556, model fields that are json data are removed from the working copy because `$this->getAttribute($field->handle())` returns `null` when the handle is `foo->bar`.

Instead we need to get `foo`.

No tests yet @duncanmcclean but this code does work.


duncanmcclean (approved):
Thanks!

PR #583, opened 2024-08-13
[7.x] Add runway:count tag

vJoeyz:
I couldn't find it anywhere in the docs or code. If it already exists or there's another way to quickly count all resources feel free to close. Inspired by the collection:count [tag](https://statamic.dev/tags/collection-count).

```php
Statamic::tag('runway:count')->in('plant')
// or
{{ runway:count in="plant" }}
```

### DIFF
--- a/docs/templating.md
+++ b/docs/templating.md
@@ -129,6 +129,20 @@ As [with the collection tag](https://statamic.dev/tags/collection#scope), you ma
 {{ /runway:post }}
 ```
 
+## Count Tag
+
+When you just want to know how many results you have, you can use the `{{ runway:count }}` tag.
+
+```antlers
+{{ runway:count from="posts" }}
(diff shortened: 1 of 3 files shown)


duncanmcclean (commented):
Would you be able to write a test for this? You can add it to the [`RunwayTagTest`](https://github.com/statamic-rad-pack/runway/blob/7.x/tests/Tags/RunwayTagTest.php) file.

duncanmcclean (approved):
Thanks for this pull request!

I've made some changes to the `RunwayTag` class so you can use the `where` and `status` parameters alongside the `{{ runway:count }}` tag, like this:

```antlers
{{ runway:count from="posts" where="author_name:duncan" }}
```

I've also added a test & mentioned it on the docs.

PR #597, opened 2024-09-09
[7.x] Add Laravel Scout and Algolia explanation to search.md

bastihilger:
I got bitten a bit by trying to use Laravel Scout in parallel with the Statamic and Runway search indexes. This addition to the docs explains what I had to do make it work.


bastihilger:
This is not ready yet, I just found out. Problems with using the search outside of Statamic now...

PR #599, opened 2024-09-16
[7.x] Return unlink when parent is not set

### LINKED ISSUE #598: Call to a member function {$relationshipName}() on null
### Description

I have two eloquent models: Scheme and Round. A Scheme has many Rounds:

```
public function rounds() {
        return $this->hasMany( Round::class );
    }
```

In the scheme.yaml blueprint, I have created a has_many field for the rounds:

```
  -
    handle: rounds
    field:
      type: has_many
      resource: round
      display: Rounds
      create: false
```

When I go to the scheme creation page in the control panel (/cp/runway/scheme/create), I get this error:

```
Call to a member function rounds() on null
```

In vendor / statamic-rad-pack / runway / src / Fieldtypes / BaseFieldtype.php : 122

I have tried the same with other models and the same error occurs. When debugging the code I can see that `$this->field->parent()` is `null` BaseFieldtype.php : 122

### Steps to reproduce

1. Create two models with a has many relation
2. Create the blueprint for model a with a field for the relation
3. Configure the model as a runway resource
4. Go to the creation page

### Environment

Environment
Application Name: XXX
Laravel Version: 10.48.14
PHP Version: 8.2.22
Composer Version: 2.3.5
Environment: local
Debug Mode: ENABLED
URL: xxx.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 1
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.16.0 PRO

Statamic Addons
statamic-rad-pack/runway: 7.9.0

BobWez98:
This might fix #598 . When i add this check it returns the page without errors and the runway resource that does not exist anymore will be marked red for the user to delete.

### DIFF
--- a/src/Fieldtypes/BaseFieldtype.php
+++ b/src/Fieldtypes/BaseFieldtype.php
@@ -114,7 +114,7 @@ class BaseFieldtype extends Relationship
 
     private function getUnlinkBehavior(): string
     {
-        if ($this instanceof BelongsToFieldtype || !$this->field->parent()) {
+        if ($this instanceof BelongsToFieldtype || ! $this->field->parent()) {
             return 'unlink';
         }
 
--- a/src/Fieldtypes/BaseFieldtype.php
+++ b/src/Fieldtypes/BaseFieldtype.php
@@ -114,7 +114,7 @@ class BaseFieldtype extends Relationship
 
     private function getUnlinkBehavior(): string
     {
-        if ($this instanceof BelongsToFieldtype) {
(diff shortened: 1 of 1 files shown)


duncanmcclean (approved):
Thanks!

PR #619, opened 2024-09-30
[7.x] Minor optimisations on isJson function

indykoning:
Profiling i noticed (in my case) jsJson is called over 10k times in the code, causing wall time to be quite high.
On large strings `json_decode` is a heavy operation, if we perform some simpler string checks we can hopefully reduce some json_decode operations.

once https://www.php.net/manual/en/function.json-validate.php can be used this can further optimise the check since it's what json_decode uses before actually attempting to decode, thus we can skip the decoding itself.

### DIFF
--- a/src/Support/Json.php
+++ b/src/Support/Json.php
@@ -2,15 +2,21 @@
 
 namespace StatamicRadPack\Runway\Support;
 
+use Illuminate\Support\Str;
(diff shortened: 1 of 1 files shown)


duncanmcclean (approved):
Thanks!

PR #626, opened 2024-10-24
[7.x] Add check whether value is null

indykoning:
I've seen a couple warnings pop up during development

> LOG.warning: json_decode(): Passing null to parameter #1 ($json) of type string is deprecated in vendor/doublethreedigital/runway/src/Support/Json.php on line 18

So we should probably add a check whether we're dealing with a null value as well


duncanmcclean (approved):
Thanks!

PR #647, opened 2025-01-02
[7.x] Check if relationship method exists before calling it

BobWez98:
In our situation we have a field "products" in a page builder fieldset, it is a HasMany field that links to the Product runway model. However there's no relationship set on the models so we're getting the error "called to undefined method "products". This way the field will check if the relationship exists before calling the method.


duncanmcclean (approved):
Thanks!

PR #654, opened 2025-02-03
[8.x] Fix attributes returning a collection instead of an array

kailumworkhouse:
Following on from https://github.com/statamic/cms/issues/11333

Fix implemented through runway.

### DIFF
--- a/src/Http/Controllers/CP/Traits/ExtractsFromModelFields.php
+++ b/src/Http/Controllers/CP/Traits/ExtractsFromModelFields.php
@@ -17,7 +17,7 @@ trait ExtractsFromModelFields
         $fields = $blueprint
             ->fields()
             ->setParent($model)
-            ->addValues($values->all())
+            ->addValues($values->toArray())
             ->preProcess();
 
         $values = $fields->values()->merge([


duncanmcclean:
Are you able to provide some more details about the issue you're running into? Changing `->all()` to `->toArray()` could potentially impact performance.

src/Http/Controllers/CP/Traits/ExtractsFromModelFields.php
```diff
@@ -17,7 +17,7 @@ protected function extractFromFields(Model $model, Resource $resource, Blueprint
         $fields = $blueprint
             ->fields()
             ->setParent($model)
-            ->addValues($values->all())
+            ->addValues($values->toArray())
```
duncanmcclean, line 20: Are you able to provide some more context about the issue you're running into and why this fixes it? Changing `->all()` to `->toArray()` may impact performance.

duncanmcclean (changes requested):

src/Http/Controllers/CP/Traits/ExtractsFromModelFields.php
simonworkhouse: Calling `$values->toArray()` will also cast any items to arrays if they are `Arrayable`, whereas calling `->all()` will simply return the items. See `Illuminate\Support\Traits\EnumeratesValues`:

```php
/**
 * Get the collection of items as a plain array.
 *
 * @return array<TKey, mixed>
 */
public function toArray()
{
    return $this->map(fn ($value) => $value instanceof Arrayable ? $value->toArray() : $value)->all();
}
```

simonworkhouse: Sorry, I forgot to mention what this fixes. It will convert a collection of strings into an array as expected by the `Statamic\Fieldtypes\HasSelectOptions` trait. An alternative fix would be this PR https://github.com/statamic/cms/pull/11332

duncanmcclean (approved):

duncanmcclean:
Rolling this back as its caused a breaking change. 

You can likely workaround the issue you were seeing by changing the casts on your model to `json` or `array`, instead of (presumably) `collection`.

simonworkhouse:
@duncanmcclean Unfortunately we cannot change the cast types in our use-case, but in my opinion the issue really lies within `statamic/cms`. I'd suggest that it's incorrect behaviour for the Statamic select field type to wrap collections in an array, see https://github.com/statamic/cms/issues/11333

Do you think that we could re-open that issue?

duncanmcclean:
Are you able to open an issue on the Runway repo, with an example repo demonstrating the issue and I can take a look when I have a chance.

I have a _feeling_ it's probably something that should be fixed on the Runway side (I might be wrong), but we just need to figure out a way to fix it without breaking relationships for others.

PR #659, opened 2025-02-10
[8.x] Improve compatibility with per model database connections

godismyjudge95:
Currently the resource gathers the columns using the Schema facade, but this facade defaults to the default database connection for the entire project.

This can lead to problems where you need to specify the connection on a per model basis - like when using the Orbit driver - https://github.com/ryangjchandler/orbit

This PR resolves this by getting columns from schema builder on the model's connection.
Everything should work just the same.  :)


duncanmcclean (approved):
Thanks!

PR #661, opened 2025-02-19
[8.x] Add withoutGlobalScopes to Rebuild URI's command

godismyjudge95:
Currently, models that have global model scope(s) set up (such as for private or hidden models) won't have their URI rebuilt/cached when this command is run.

This PR adds a simple `->withoutGlobalScopes()` to bypass the scopes while the cache is being rebuilt.

I don't think there are any long term implications against this, but let me know if there are.  Could also just place it behind an option flag if so.


duncanmcclean (approved):
Thanks!

PR #663, opened 2025-02-21
[8.x] Fix disappearing status filter on collections 

### LINKED ISSUE #662: Status filter not visible to Statamic collection listing
### Description

The status filter is missing in listings for Statamic collections, when runway is installed.

I noticed that `StatamicRadPack\Runway\Scopes\Status` class overrides the `visibleTo()` method of `Statamic\Query\Scopes\Filters\Status,` which is why the filter is missing. 

```php
public function visibleTo($key): bool
{
    return in_array($key, ['runway']) && $this->resource()->hasPublishStates();
}
```

A possible quick solution to this problem could be to determine the context and call the method from the correct class. This would of course also be necessary for the `options()` method in the very same class.

```php
use Statamic\Query\Scopes\Filters\Status as BaseStatusFilter;

class Status extends BaseStatusFilter
{

    public function visibleTo($key): bool
    {
      if ($this->shouldUseBaseStatusFilter()) {
          return BaseStatusFilter::visibleTo($key);
      }
    
      return in_array($key, ['runway']) && $this->resource()->hasPublishStates();
    }

    public function options(): Collection
    {
      if ($this->shouldUseBaseStatusFilter()) {
          return BaseStatusFilter::options();
      }
  
      return collect([
          'published' => __('Published'),
          'draft' => __('Draft'),
      ]);
    }
    
    protected function shouldUseBaseStatusFilter(): bool
    {
       return isset($this->context['collection']);
    }
}
```

### Steps to reproduce

1. Install runway package
2. Create a Statamic collection
3. Go to collection listing
4. The `Status` filter is not visible

Image

### Environment

Environment
Laravel Version: 11.41.3
PHP Version: 8.4.2
Composer Version: -
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: UTC
Locale: de

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: mysql
Logs: stack / single
Mail: log
Queue: database
Session: file

Livewire
Livewire: v3.5.19

Statamic
Addons: 3
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.46.1 PRO

Statamic Addons
statamic-rad-pack/runway: 7.14.0
visuellverstehen/statamic-classify: 3.0.0
visuellverstehen/statamic-picturesque: 1.4.7

duncanmcclean:
This pull request fixes an issue where Runway would override the "Status" filter from Statamic, preventing it from showing on collection listings.

Fixes #662.

### DIFF
--- a/src/Scopes/Status.php
+++ b/src/Scopes/Status.php
@@ -9,6 +9,8 @@ use StatamicRadPack\Runway\Runway;
 
 class Status extends BaseStatusFilter
 {
+    protected static $handle = 'runway-status';
+
     public function visibleTo($key): bool
     {
         return in_array($key, ['runway']) && $this->resource()->hasPublishStates();


duncanmcclean:
No problem!

PR #705, opened 2025-08-01
[8.x] Allow programmatic registration of resources

Pluuk:
This PR enables Statamic addons to register their own resources to Statamic. This enables resources to be more dynamic and don't require resources to be manually added to the config each time an addon is registrered.

The `registerResource` function should be called in the register function the addon serviceprovider.

### DIFF
--- a/src/Runway.php
+++ b/src/Runway.php
@@ -14,27 +14,20 @@ class Runway
 
     public static function discoverResources(): self
     {
-        static::$resources = collect(config('runway.resources'))
+        static::$resources = array_merge(static::$resources, collect(config('runway.resources'))
             ->mapWithKeys(function ($config, $model) {
-                $config = collect($config);
-                $handle = $config->get('handle', Str::snake(class_basename($model)));
+                $resource = static::buildResource($model, $config);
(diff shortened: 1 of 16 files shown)


Pluuk:
The error that is thrown in the Test Suite is the following:
```
InvalidArgumentException: Resource with handle post already registered
```

I guess this is a faulty test, because the registered resources are registered multiple times. Not sure what the best course of action is. Should be allow the resources to be overridden or not? It also feels a bit weird if it should be allowed, because having duplicate handles can cause unexpected behaviour.

Pluuk:
I'm creating an addon which I want to use for multiple projects and this addon needs some data storage that the user can perform CRUD operations on. I don't want to just publish the collection files or handle it with configuration because that's manual work. I want it to be plug and play.

duncanmcclean:
After reviewing your changes, I was able to come up with a much simpler approach which you can find in #707. I wasn't able to push to your fork, so I've opened a new PR for the changes. 

Thanks for the PR! 😄

Pluuk:
Looking good, thanks!

PR #706, opened 2025-08-08
[8.x] Update RoutingModel model property to be publicly accessible

jamesmpigott:
This PR adds a `resource` method on the `RoutingModel` class, allowing for easy access to the Runway Resource from the `page` variable on a runway route. I've opted to leave the `model` property as protected (as there's probably a reason for that choice).

I've also refactored the `template` and `layout` methods to utilise the new `resource` method, for consistency. Tests for these methods are passing.

Also adds a test for the new `resource` method.

For more context, please see Discord thread: https://discord.com/channels/489818810157891584/1151898550305837086/1403406370275065918

^ If the Discord link above is inaccessible, I'm building a admin toolbar (like WordPress) that's visible on the frontend, and allows the user to edit the entry they're currently on, amongst other things. I'd got this working with Statamic Collections, but the need has arisen to support Runway resources.

### DIFF
--- a/src/Routing/RoutingModel.php
+++ b/src/Routing/RoutingModel.php
@@ -2,12 +2,13 @@
 
 namespace StatamicRadPack\Runway\Routing;
 
-use Illuminate\Contracts\Support\Responsable;
+use StatamicRadPack\Runway\Runway;
+use Statamic\Data\HasAugmentedData;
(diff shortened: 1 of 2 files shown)


duncanmcclean (approved):
Thanks James! 🙂

PR #711, opened 2025-08-14
[8.x] fix: inline actions on edit view #710

### LINKED ISSUE #710: Only super admins can use inline action on edit resource view
### Description

When editing a resource as a non super admin, you cannot see the three dots where the actions for the resources are listed. This is because the PublishForm Vue component (/resources/js/components/resources/PublishForm.vue) is using a v-if on line 13 and checks for the "canEditBlueprint" permission on the whole actions dropdown.

### Steps to reproduce

1. Create a Resource
2. Create a Statamic User that is not super admin
3. Create a Statamic action
4. On the list view of the resource the non super admin can see the actions
5. On the edit view of the resource the non super admin cannot see the actions 

### Environment

Environment
Application Name: redacted
Laravel Version: 11.44.7
PHP Version: 8.4.11
Composer Version: 2.8.10
Environment: local
Debug Mode: ENABLED
URL: redacted
Maintenance Mode: OFF
Timezone: UTC
Locale: de

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: mysql
Logs: stack / single
Mail: smtp
Queue: database
Session: file

Livewire
Livewire: v3.6.3

Statamic
Addons: 3
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.54.0 PRO

Statamic Addons
statamic-rad-pack/runway: 8.5.1
visuellverstehen/statamic-classify: 3.0.0
visuellverstehen/statamic-picturesque: 1.4.7

persteinhorst:
fixes #710

### DIFF
--- a/resources/js/components/resources/PublishForm.vue
+++ b/resources/js/components/resources/PublishForm.vue
@@ -10,8 +10,8 @@
                 </div>
             </h1>
 
-            <dropdown-list class="rtl:ml-4 ltr:mr-4" v-if="canEditBlueprint">
-                <dropdown-item :text="__('Edit Blueprint')" :redirect="actions.editBlueprint" />
+            <dropdown-list class="rtl:ml-4 ltr:mr-4">
+                <dropdown-item v-if="canEditBlueprint" :text="__('Edit Blueprint')" :redirect="actions.editBlueprint" />
                 <li class="divider" />
                 <data-list-inline-actions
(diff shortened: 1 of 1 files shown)


duncanmcclean (approved):
Thanks!

PR #714, opened 2025-08-20
[8.x] Add column existence validation for Runway resource fields

kevinmeijer97:
We've implemented functionality in our https://github.com/rapidez/statamic package that adds a hidden collection to a Runway resource, including the collection's fields in the resource's blueprint.

However, we encountered an issue when adding new fields to the blueprint with names that don't correspond to existing columns in the resource's database table. This correctly throws an error since the column doesn't exist.

This PR adds validation to ensure that all fields being searched against the resource table have corresponding columns in the database.

### DIFF
--- a/src/Traits/HasRunwayResource.php
+++ b/src/Traits/HasRunwayResource.php
@@ -3,6 +3,7 @@
 namespace StatamicRadPack\Runway\Traits;
 
 use Illuminate\Contracts\Database\Eloquent\Builder;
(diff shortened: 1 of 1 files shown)


duncanmcclean (approved):
Thanks!

Now that we're checking if the column exists, we probably don't need the `instanceof` checks at all since columns won't exist for HasMany/Hidden/Section fields. I've simplified the logic.

PR #715, opened 2025-08-23
Fix runway:import-collection to run on Ubuntu, and not just Mac

amadeann:
`spl_autoload` function maps classes to lowercase filenames:
https://www.php.net/manual/en/function.spl-autoload.php
https://stackoverflow.com/a/14361521/4171578

When you run `runway:import-collection` on Mac, it works fine, since Mac filesystem is case-insensitive. On Ubuntu, running this throws an exception like:

```
class_parents(): Class App\Models\Post does not exist and could not be loaded
```

I'm running Laravel Sail in WSL2 on Windows. Changing `spl_autoload` to `require_once` fixes the issue.

### DIFF
--- a/src/Console/Commands/ImportCollection.php
+++ b/src/Console/Commands/ImportCollection.php
@@ -176,7 +176,7 @@ PHP);
 
         File::put(app_path("Models/{$this->modelName}.php"), $modelContents);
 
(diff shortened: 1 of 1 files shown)


duncanmcclean (approved):
Thanks!

PR #733, opened 2025-10-21
Fixes `published() on null` error if model missing

joshtaylordev:
Fixes the `Call to a member function published() on null` error.

This occurs when a model has a published column and is referenced on a page using the `belongsTo` or `hasMany` fieldtypes, and the model doesn't exist (e.g. it's been deleted).


duncanmcclean:
Thanks!

PR #746, opened 2025-11-12
Handle enums correctly in listings

janis-ps:
Problem
When Eloquent casts (e.g., backed enums) are applied to a model attribute that is managed by a Statamic Select field, Runway currently passes the casted value (e.g., an enum instance) into the fieldtype's preProcessIndex(). Because Select options are defined as scalar values (strings/ints), the label lookup fails in listings and Statamic shows the raw option value instead of the human-friendly label.

Solution
Introduce a per-resource config option ignore_casts (array of field handles). For fields listed here, Runway will feed the raw persisted value to preProcessIndex() by calling getRawOriginal($key) instead of getAttribute($key). This preserves the scalar value so the Select fieldtype can resolve its label correctly.

Changes
- Add Resource::ignoreCasts(string $field)
- In Http\Resources\CP\ListedModel::values(), use getRawOriginal() when ignoreCasts($key) is true; otherwise continue using getAttribute().

Config example
resources:
  your_resource_handle:
    ignore_casts: ['status']

Backward compatibility
- Opt-in only; default behavior is unchanged.

Why
We have columns cast to backed enums. When those columns are managed by Select fields, everything works except in listings: Statamic shows the option value instead of the human-friendly label since Runway passes the enum object, which cannot match any of the Select option values.

Notes
- This specifically improves the interaction with Statamic's Select fieldtype (HasSelectOptions) which expects scalar values when deriving labels in index views.

AI-generated content disclosure
This PR (proposal, commit message, and patch) was prepared with assistance from an AI tool (Augment Agent, based on OpenAI GPT). Please review carefully; any mistakes are mine and I'll happily make requested changes.

---
Pull Request opened by [Augment Code](https://www.augmentcode.com/) with guidance from the PR author

### DIFF
--- a/tests/__fixtures__/database/migrations/2020_12_12_220453_create_posts_table.php
+++ b/tests/__fixtures__/database/migrations/2020_12_12_220453_create_posts_table.php
@@ -3,7 +3,6 @@
 use Illuminate\Database\Migrations\Migration;
 use Illuminate\Database\Schema\Blueprint;
 use Illuminate\Support\Facades\Schema;
-use StatamicRadPack\Runway\Tests\Fixtures\Enums\MembershipStatus;
 
 class CreatePostsTable extends Migration
 {


duncanmcclean:
What's your use case? Enums?

janis-ps:
Yes, we have an `Article` model (and many others) that have a `status` attribute that we cast to an enum via Eloquent. But  when we define that attribute as a Select field in a blueprint and show it in control panel listings, Statamic ends up not finding the label to show (Statamic finds labels with `===` and a string will never match an enum instance).

PR #753, opened 2025-11-24
[8.x] Fix runway search on models with individual connections

godismyjudge95:
We recently ran into an issue where the search on our Runway models was not working properly (as in it wasn't filtering at all):  https://discord.com/channels/489818810157891584/1151898550305837086/1415710334219583518

After a bit of digging, I found out that it is because we are using the Orbit Eloquent driver for the model, and Orbit specifies the DB connection on a per model basis.  This means that when this code is called for scoping the Runway search:
```php
    public function scopeRunwaySearch(Builder $query, string $searchQuery)
    {
        $this->runwayResource()->blueprint()->fields()->all()
            ->filter(fn (Field $field) => Schema::hasColumn($this->getTable(), $field->handle()))
            ->reject(fn (Field $field) => $field->visibility() === 'computed')
            ->each(fn (Field $field) => $query->orWhere($this->getColumnForField($field->handle()), 'LIKE', '%'.$searchQuery.'%'));
    }
```

it gets the default DB connection and not the Orbit connection.

This is not an Orbit specific issue as this behavior would happen with any model that specifies a different DB connection than the default.

This PR adds a failing test and a simple fix to use the model's connection and subsequently schema builder to check for the columns' existence.

### DIFF
--- a/src/Runway.php
+++ b/src/Runway.php
@@ -77,6 +77,13 @@ class Runway
         return new static;
     }
 
+    public static function clearRegisteredResources(): self
+    {
+        static::$registeredResources = [];
+
+        return new static;
+    }
+
     public static function usesRouting(): bool
     {
         return static::allResources()->filter->hasRouting()->count() >= 1;
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -8,6 +8,7 @@ use Statamic\Facades\Blueprint;
 use Statamic\Stache\Stores\UsersStore;
 use Statamic\Statamic;
 use Statamic\Testing\AddonTestCase;
+use StatamicRadPack\Runway\Runway;
 use StatamicRadPack\Runway\ServiceProvider;
 
 abstract class TestCase extends AddonTestCase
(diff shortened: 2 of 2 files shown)


tests/__fixtures__/app/Models/ExternalPost.php
```diff
@@ -0,0 +1,20 @@
+<?php
```
duncanmcclean: We're only using this model in the `ResourceTest` file. To keep things self-contained, let's move this to the bottom of that file.

tests/ResourceTest.php
```diff
@@ -301,4 +303,45 @@ public function revisions_cant_be_enabled_without_publish_states()
...
+        Schema::connection('external')->create('external_posts', function ($table) {
+            $table->id();
+            $table->string('title');
+            $table->longText('body');
+            $table->timestamps();
+        });
+
+        Config::set('runway.resources.'.ExternalPost::class, []);
+
+        Runway::discoverResources();
```
duncanmcclean: Let's call `registerResource` here instead. 

It avoids us needing to set the config, then tell Runway to find all the resources again. It's a tiny bit cleaner IMO.

```suggestion
        Runway::registerResource(ExternalPost::class, []);
```

godismyjudge95:
Failing tests appear to be Github actions having issues again... 🫠
Implemented your suggestions.

PR #785, opened 2026-03-09
[9.x] Performance: prevent routingModel reinit

macaws:
This small change prevents needless class reinstantiation


duncanmcclean:
Thanks!

PR #793, opened 2026-03-25
[9.x] Fix error on listing page when using model attribute as first column in blueprint

### LINKED ISSUE #759: Resource Listing sorting error when using computed field
### Description

When using a Title attribute on a resource model, a `Unknown column 'title' in 'order clause'` error is thrown when viewing the resource listing. (/cp/runway/resource)

I tried setting `order_by` in the resource config, however it looks like the initial sort field (`$request->input('sort')`) uses the first field in the blueprint (`title`). 

```
\App\Models\Response::class => [
    'name' => 'Responses',
    'cp_icon' => 'drawer-file',
    'order_by' => 'created_at',
    'title_field' => 'title',
],
```

A workaround for this is simply moving my `created_at` field before the `title` field in my blueprint.

```
        fields:
          -
            handle: created_at
            field:
              type: date
              display: Date
              listable: true
              default: now
              validate:
                - required
              format: 'Y-m-d H:i:s'
              visibility: read_only
          -
            handle: title
            field:
              type: text
              display: Title
              visibility: computed
              sortable: false
              listable: true
```

_Note: This error returns when using a MySQL db. SQLite appears to be more forgiving and doesn't throw an error, hence why I only had this issue on my staging server._

Image

### Steps to reproduce

1. Use MySQL db
2. Migrate a resource with no Title column
3. Add a Title attribute in the resource model
4. View resource listing in CP

### Environment

Environment
Application Name: Change Helm
Laravel Version: 12.44.0
PHP Version: 8.3.27
Composer Version: 2.8.11
Environment: local
Debug Mode: ENABLED
URL: changehelm.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Livewire
Livewire: v3.6.4

Statamic
Addons: 9
Sites: 1
Stache Watcher: Enabled (auto)
Static Caching: Disabled
Version: 6.0.0-alpha.20 PRO

Statamic Addons
jacksleight/statamic-bard-mutator: 3.0.5
jacksleight/statamic-bard-texstyle: dev-feature/statamic-6
marcorieser/statamic-livewire: dev-statamic-6
marcorieser/tailwind-merge-statamic: dev-statamic-6
statamic-rad-pack/runway: 9.0.0-alpha.4
studio1902/statamic-peak-browser-appearance: dev-statamic-6
studio1902/statamic-peak-commands: 2.2
studio1902/statamic-peak-seo: dev-statamic-6
studio1902/statamic-peak-tools: dev-statamic-6

andjsch:
This PR adds two new methods to the ``HasRunwayResource`` trait for controlling the default sort column and direction used within the Listing UI component. By default, the component would fall back to the first column of the blueprint, which results in an error if the column is computed. 

By default the newly added ``sortColumn()`` and ``sortDirection()`` methods are falling back to the primary key of the model and ``asc`` which should be sensible defaults and mimic the prior behaviour of version 8.x 

fixes #759


duncanmcclean (approved):
Thanks for the pull request! 🙌

I've changed the implementation slightly to use the existing [`order_by` and `order_by_direction`](https://runway.duncanmcclean.com/resources#ordering) config options.

PR #806, opened 2026-04-13
[9.x] Qualify the primary key in queries

Jade-GG:
This fixes an edge case that occurs when you have a resource where querying it requires disambiguating (qualifying) the primary key. For example, the specific case we ran into was using product attributes from Magento. Within our code, we join the `eav_attribute` table onto the `catalog_eav_attribute` table. The primary key of the `eav_attribute` table is called `attribute_id`, which is also the column name they use for the foreign key in `catalog_eav_attribute`.

This is (unfortunately) a rather common occurrence in Magento so we often already have to qualify columns everywhere. However, we run into an issue with Runway because it currently assumes primary keys don't need to be qualified when it does queries.

---

I couldn't change the `primaryKey()` function itself as it also gets used outside of queries, so instead I made a new `qualifiedPrimaryKey()` function. I think I caught all of the usages of the function here that relate to querying.

Note: I have not written a test for this right now, as it would require modifying the testing database/migrations structure to fit this specific edge case. That seemed overkill, and frankly just a little bit out of my wheelhouse 😅

### DIFF
--- a/src/Fieldtypes/BaseFieldtype.php
+++ b/src/Fieldtypes/BaseFieldtype.php
@@ -293,7 +293,7 @@ abstract class BaseFieldtype extends Relationship
                     $item = $item->first()->{$resource->primaryKey()} ?? null;
                 }
 
-                $model = $resource->model()->firstWhere($resource->primaryKey(), $item);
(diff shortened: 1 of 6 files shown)


duncanmcclean (approved):
Thanks!

PR #822, opened 2026-05-26
[9.x] Fix date filter `between` operator not working in runway listing

albinameti00:
## Problem
When filtering a Runway resource listing by a date field using the `between` operator,
the filter is silently ignored and all results are returned unfiltered.

This is because `isComplete()` only checks for the `value` key, but the `between`
operator sends `range_value` instead:

- `after` / `before` → `{ "operator": ">", "value": "..." }`
- `between` → `{ "operator": "between", "range_value": { "start": "...", "end": "..." } }`

So `isComplete()` always returns `false` for `between`, preventing the filter from ever being applied.

## Fix
Add a `between` case to `isComplete()` that checks for `range_value.start` and `range_value.end`.

### DIFF
--- a/src/Scopes/Fields.php
+++ b/src/Scopes/Fields.php
@@ -59,7 +59,7 @@ class Fields extends BaseFieldsFilter
         }
 
         if ($operator === 'between') {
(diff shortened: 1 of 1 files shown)


duncanmcclean:
Thanks!

PR #823, opened 2026-05-27
[9.x] Fix belongs_to not being populated during inline has_many creation

### LINKED ISSUE #799: `has_many` field doesn't let inline creation if foreign key is not nullable
### Description

I have two models, let's say Folder and File.
Folder has many Files, and File belongs to folder.

I've set these so that the file must have a folder (folder_id foreign key set as non nullable).

I go and create a folder, then with the has_many fieldtype I go and create the file. When I try to save the file I get the error `folder_id doesn't have a default value` because it's null.

The current flow as I understood it is:
- When you save the child you save it as is without the reference to the parent.
- The reference to the parent gets saved in the child element only when you save the parent as well.
- When creating the child you cannot save both, hence the parent id must be nullable to save it.

The current workaround I found is to make the foreign keys nullable, but then you cannot have the option to delete the children, as the only option that's there now is to unlink them.

### Steps to reproduce

1. Create some models with the same relationship as the example (with non nullable foreign kets)
2. Add them to as a runway resource
3. Create a blueprint for both, with an has_many fieldtype in the parent and a belongs_to fieldtype on the child.
4. Try to create a child from a saved parent.

### Environment

Environment
Laravel Version: 12.56.0
PHP Version: 8.4.19
Composer Version: 2.8.4
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: UTC
Locale: it

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 2
Sites: 1
Stache Watcher: Enabled (auto)
Static Caching: Disabled
Version: 6.11.0 Solo

Statamic Addons
statamic-rad-pack/runway: 9.3.3
statamic/eloquent-driver: 5.6.2

Statamic Eloquent Driver
Addon Settings: file
Asset Containers: file
Assets: eloquent
Blueprints: file
Collection Trees: file
Collections: file
Entries: eloquent
Fieldsets: file
Form Submissions: eloquent
Forms: file
Global Sets: file
Global Variables: file
Navigation Trees: file
Navigations: file
Revisions: file
Sites: file
Taxonomies: file
Terms: file
Tokens: eloquent

macaws:
Goal is to fix https://github.com/statamic-rad-pack/runway/issues/799

### DIFF
--- a/resources/js/components/PublishForm.vue
+++ b/resources/js/components/PublishForm.vue
@@ -474,7 +474,7 @@ export default {
         },
 
         parentContainer() {
-            return this.publishContext?.parentContainer;
+            return this.publishContext;
         },
     },
 


duncanmcclean (approved):
Thanks!

PR #840, opened 2026-07-22
[9.x] Batch relationship fieldtype augmentation into a single query

ryanmitchell:
BaseFieldtype::getAugmentableModels() resolved each id in a has_many/ relationship field one at a time via firstWhere(), so a field referencing N models triggered N queries (plus N more per eager-loaded relation) every time it was augmented for the front end.

Batch the ids that aren't already Blink-cached into a single whereIn()->with(...) query, populate Blink per-model afterwards, then read every value back out of Blink in original order. Blink caching behaviour (skipping ids already resolved elsewhere in the same request) and the with()/eagerLoadingRelationships() config are unchanged.


duncanmcclean (approved):

duncanmcclean:
Thanks!

PR #850, opened 2026-09-03
[9.x] Configuration option disable static warm

mefenlon:
This pull request adds a configuration option to disable warming of runway's uris.

When using Statamic's [warming additional URLs](https://statamic.dev/advanced-topics/static-caching#warming-additional-urls) feature, developers may want complete control over which URLs are warmed. This option allows Runway URI warming to be disabled, preventing possible overlap with custom warming configurations and enabling fully customized cache warming strategies.

### DIFF
--- a/config/runway.php
+++ b/config/runway.php
@@ -42,7 +42,6 @@ return [
 
     'disable_migrations' => false,
 
-
     /*
     |--------------------------------------------------------------------------
     | Disable Static Warm?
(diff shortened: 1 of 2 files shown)


duncanmcclean (approved):
Thanks for the PR, @mefenlon!

I've renamed the config option to `static_warming` as negative booleans get confusing quickly (I know `disable_migrations` is like that... but it's old 😅)

```php
'static_warming' => true,
```

The `ServiceProvider` will fallback to `true` when the key is missing, so it'll continue to warm URIs on existing apps.

